PR #2, opened 2025-12-09
Replace table for trailing zeros with direct calculation.

TobiSchluter:
żmij reads the number of trailing zeros from a table in order to adjust the end of buffer. Determining the number of trailing zeros in the four-byte blocks written can be done in three or four assembly instructions and doing it this way performs equally well in my tests with MSVC. Tested by copying zmij.cc into dtoa-benchmark and running it. On my laptop the current trunk and this version both fluctuate between 25ns and 26ns per conversion.

A big-endian architecture (if it exists) would have to search for the non-zero bits in the opposite direction.

In my builds the dtoa-benchmark executable shrinks by 512 bytes, from 235008 bytes to 234496 bytes, i.e. 512 bytes less. The .obj itself gets larger, but I haven't studied the reason. 

It's been fun following your adventures into schubfach, which i can pronounce (and type!) with much more confidence than żmij.

Edit: as usual I got little-endian and big-endian mixed up, and fixed the comment mentioning it with a forced push.

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -705,14 +705,6 @@ inline auto digits2(size_t value) noexcept -> const char* {
   return &data[value * 2];
 }
 
-// The idea of branchless trailing zero removal is by Alexander Bolz.
-const char num_trailing_zeros[] =
-    "\2\0\0\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0\0\0"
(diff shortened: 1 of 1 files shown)


TobiSchluter:
I noticed that the bbcc case misses -4 on line 764 when addressing the buffer, and so I want to recheck the correctness a bit more carefully.

PR #3, opened 2025-12-09
Replace table for trailing zeros with direct calculation.

TobiSchluter:
Edit: I installed WSL and played around it on Linux. I learnt that the behavior of __builtin_clz chagnes with compiler flags (undefined vs 32 for zero argument). The workarounds I could come up with for this seemed to pessimize performance (increase to ~28ns), so I don't think this is worth pursuing. The amount of space this saves is minimal after all, and performance (in the cases where it works correctly, i.e. windows or -mlzcnt) is the same as the extant implementation.

żmij reads the number of trailing zeros from a table in order to adjust the end of buffer. Determining the number of trailing zeros in the four-byte blocks written can be done in three or four assembly instructions and doing it this way performs equally well in my tests with MSVC. Tested by copying zmij.cc into dtoa-benchmark and running it. On my laptop the current trunk and this version both fluctuate between 25ns and 26ns per conversion. I tried using `std::countl_zero` but at least with msvc it is significantly slower (timings flluctuate between 27ns and 30ns if I use it), so I settled on the compiler intrinsics.

A big-endian architecture (if it exists) would have to search for the non-zero bits in the opposite direction.

In my builds the dtoa-benchmark executable shrinks by 512 bytes, from 235008 bytes to 234496 bytes, i.e. 512 bytes less. The .obj itself gets larger, but I haven't studied the reason.

It's been fun following your adventures into schubfach, which i can pronounce (and type!) with much more confidence than żmij.

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -705,14 +705,6 @@ inline auto digits2(size_t value) noexcept -> const char* {
   return &data[value * 2];
 }
 
-// The idea of branchless trailing zero removal is by Alexander Bolz.
-const char num_trailing_zeros[] =
-    "\2\0\0\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0\0\0"
-    "\1\0\0\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0\0\0"
-    "\1\0\0\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0\0\0"
-    "\1\0\0\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0\0\0"
-    "\1\0\0\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0\0\0";
-
 struct div_mod_result {
   uint32_t div;
   uint32_t mod;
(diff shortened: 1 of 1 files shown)


vitaut:
Thank you for the PR! That's a very clever idea but you are correct in the assessment that it pessimizes performance a bit. So I agree that it's not pursuing at this time but we might look into it later if we decide to provide a version optimized for binary size. The first step would be to compress powers of 10 though.

TobiSchluter:
@vitaut thanks for the encouragement. It made me look at this a bit closer, and I realized that the approach can be taken a bit further which allows getting rid of two if statements at the expense of two now unconditional (fast) divisions. This cuts the assembly from 353 to 327 lines and it also seems to consistently be faster by 1-2ns, though the measurements are quite inaccurate.

I'll clean it up, and send a new pull request, hopefully tomorrow.

foonathan:
>  learnt that the behavior of __builtin_clz chagnes with compiler flags (undefined vs 32 for zero argument).

Have you tried using `__builtin_clzg(x, 32)`? It gives the value of the second argument for `x == 0`: https://gcc.gnu.org/onlinedocs/gcc/Bit-Operation-Builtins.html#index-_005f_005fbuiltin_005fclzg

TobiSchluter:
@foonathan thanks for the hint. I found that if I tell the compiler what I want by hand-coding the bit-counting in C++, it can pick the right instruction during optimization. If you look closely at the assembly from my new pull request https://github.com/vitaut/zmij/pull/4 it can actually prove that the argument is non-zero in the non-exceptional case (>8 digits) and only keeps the check in the abbreviated case.

TobiSchluter:
@foonathan can't tag you in https://github.com/vitaut/zmij/pull/41 but you had the right midterm vision, making the function robust against zero inputs is the better strategy. Thanks for pointing me to that function

PR #4, opened 2025-12-12
Write digits in groups of eight

TobiSchluter:
Writes the digits in blocks of eight instead of two and then adjust for the triailing zeros only once. They are identified by bit-counting of the eight-digit block.

This is an extension of my previous pull request. Unlike the previous, this benchmarks on par on linux and avoids all issues due to bit counting builtins not being clearly specified. This is achieved by relying on the compiler to find the right instruction. I explored `std::countl_zero` but it seemed to pessimize performance, at least on Windows.

On Windows it generally benchmarks minimally faster, on Linux it benchmarks on par, and maybe faster when `-mlzcnt` is given which allows the more efficient LZCNT instruction to be used (unlike BSR it needs no special handling of the zero case).

It removes two branches compared to the main branch, but adds two unconditional multiplications in the case of short decimal representations (which are unlikely, because a zero digit is less probable than a non-zero digit).

Here is a "lucky" benchmark run with `kTrial == 20` on Linux, generally the relative performance fluctuates a little. `zmjj_new` is this version
image

This version clocks in at <s>322</s>320 lines of assembly <s>https://godbolt.org/z/5YTzY3zEb</s>https://godbolt.org/z/f3vv1Y3f7 (updated for second patch).
The current main branch compiles to 339 lines of assembly, including an additional 100byte buffer https://godbolt.org/z/exr3E581E

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -710,13 +710,41 @@ inline auto digits2(size_t value) noexcept -> const char* {
   return &data[value * 2];
 }
 
-// The idea of branchless trailing zero removal is by Alexander Bolz.
-const char num_trailing_zeros[] =
-    "\2\0\0\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0\0\0"
-    "\1\0\0\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0\0\0"
-    "\1\0\0\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0\0\0"
-    "\1\0\0\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0\0\0"
-    "\1\0\0\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0\0\0";
+// MSVC makes lzcnt available always, but it's not constexpr
+inline auto lzcntl(uint64_t x) noexcept -> size_t
+{
+#ifdef _MSC_VER
+  return _lzcnt_u64(x);
+#else
+  // Unlike MSVC, clang and gcc recognize this implementation and replace
+  // it with the assembly instructions which are appropriate for the
+  // target (lzcnt or bsr + zero handling).
+  size_t n = 64;
+  while (x > 0) {
+    x >>= 1;
+    --n;
+  }
+  return n;
+#endif
+}
+
+inline auto count_trailing_nonzeros(uint64_t x) noexcept -> size_t {
+  // This assumes little-endian, that is the first char of the string
+  // is in the lowest byte and the last char is in the highest byte.
+  // We count the number of characters until there are only '0' == 0x30
+  // characters left.
+  // The code is equivalent to
+  //   return 8 - lzcntl(x & ~0x30303030'30303030) / 8
+  // but if the BSR instruction is emitted, subtracting the constant
+  // before dividing allows combining it with the subtraction from BSR
+  // counting in the opposite direction.
+  //   return size_t(71 - lzcntl(x & ~0x30303030'30303030)) / 8;
+  // Additionally, the bsr instruction requires a zero check.  Since the
+  // high bit is never set we can avoid the zero check by shifting the
+  // datum left by one and using XOR to both remove the 0x30s and insert
+  // a sentinel bit at the end. 
+  return size_t(70 - lzcntl(x << 1 ^ (0x30303030'30303030ull << 1 | 1))) / 8;
+}
 
 struct divmod_result {
   uint32_t div;
@@ -736,13 +764,17 @@ inline void write2digits(char* buffer, uint32_t value) noexcept {
   memcpy(buffer, digits2(value), 2);
 }
 
-// Writes 4 digits and removes trailing zeros.
-auto write4digits_trim_zeros(char* buffer, uint32_t value) noexcept -> char* {
-  auto [aa, bb] = divmod100(value);
-  write2digits(buffer + 0, aa);
-  write2digits(buffer + 2, bb);
-  return buffer + 4 - num_trailing_zeros[bb] -
-         (bb == 0) * num_trailing_zeros[aa];
+auto write8digits(uint32_t aa, uint32_t bb, uint32_t cc, uint32_t dd) noexcept -> uint64_t {
+  uint16_t bitsaa;
+  memcpy(&bitsaa, digits2(aa), 2);
+  uint16_t bitsbb;
+  memcpy(&bitsbb, digits2(bb), 2);
+  uint16_t bitscc;
+  memcpy(&bitscc, digits2(cc), 2);
+  uint16_t bitsdd;
+  memcpy(&bitsdd, digits2(dd), 2);
+
+  return uint64_t(bitsdd) << 48 | uint64_t(bitscc) << 32 | uint64_t(bitsbb) << 16 | bitsaa;
 }
 
 // Writes a significand consisting of 16 or 17 decimal digits and removes
@@ -757,28 +789,29 @@ auto write_significand(char* buffer, uint64_t value) noexcept -> char* {
   uint32_t abb = abbcc / 100;
   uint32_t cc = abbcc % 100;
   auto [a, bb] = divmod100(abb);
+  auto [dd, ee] = divmod100(ddee);
 
   *buffer = char('0' + a);
   buffer += a != 0;
-  write2digits(buffer + 0, bb);
-  write2digits(buffer + 2, cc);
-  buffer += 4;
 
-  if (ffgghhii == 0) {
-    if (ddee != 0) return write4digits_trim_zeros(buffer, ddee);
-    return buffer - num_trailing_zeros[cc] - (cc == 0) * num_trailing_zeros[bb];
+  // Use an intermediate uint64_t to make sure that the compiler constructs
+  // the value in a register.  This way the buffer is written to memory in
+  // one go and count_trailing_nonzeros doesn't have to load from memory.
+  uint64_t bits = write8digits(bb, cc, dd, ee);
+  memcpy(buffer, &bits, 8);
+
+  if (ffgghhii == 0) [[unlikely]] {
+    return buffer + count_trailing_nonzeros(bits);
   }
-  auto [dd, ee] = divmod100(ddee);
+
+  buffer += 8;
   uint32_t ffgg = ffgghhii / 10'000;
   uint32_t hhii = ffgghhii % 10'000;
   auto [ff, gg] = divmod100(ffgg);
-  write2digits(buffer + 0, dd);
-  write2digits(buffer + 2, ee);
-  write2digits(buffer + 4, ff);
-  write2digits(buffer + 6, gg);
-  if (hhii != 0) return write4digits_trim_zeros(buffer + 8, hhii);
-  return buffer + 8 - num_trailing_zeros[gg] -
-         (gg == 0) * num_trailing_zeros[ff];
+  auto [hh, ii] = divmod100(hhii);
+  uint64_t bits2 = write8digits(ff, gg, hh, ii);
+  memcpy(buffer, &bits2, 8);
+  return buffer + count_trailing_nonzeros(bits2);
 }
 
 // Writes the decimal FP number dec_sig * 10**dec_exp to buffer.
--- a/zmij-test.cc
+++ b/zmij-test.cc
@@ -10,6 +10,19 @@ auto dtoa(double value) -> std::string {
   return buffer;
 }
 
+TEST(zmij_test, utilities) {
+  EXPECT_EQ(lzcntl(0), 64);
+  EXPECT_EQ(lzcntl(1), 63);
+  EXPECT_EQ(lzcntl(~0ull), 0);
+
+  EXPECT_EQ(count_trailing_nonzeros(0x30303030'30303030ull), 0);
+  EXPECT_EQ(count_trailing_nonzeros(0x30303030'30303031ull), 1);
+  EXPECT_EQ(count_trailing_nonzeros(0x30303030'30303039ull), 1);
+  EXPECT_EQ(count_trailing_nonzeros(0x30393030'39303030ull), 7);
+  EXPECT_EQ(count_trailing_nonzeros(0x31303030'30303030ull), 8);
+  EXPECT_EQ(count_trailing_nonzeros(0x39303030'30303030ull), 8);
+}
+
 TEST(zmij_test, umul192_upper64_modified) {
   auto pow10 = pow10_significands[0];
   EXPECT_EQ(


vitaut:
Thanks for the PR! Looks like there are some merge conflicts, could you rebase?

xjb714:
@TobiSchluter I ran this code and encountered an assert failure on `assert(value < (num_digits == 3 ? 1'000 : 10'000));`

TobiSchluter:
@xjb714 i can't reproduce this with the version in the PR  though I rmeember running into that when I had accidentally written `divmod<3>` when working on a four-digit number

@vitaut I'm about to update the PR, but since I was playing with compiler options I seem to have lost native uint128_t on my Windows PC and I'm running into the assert on line 24
```
    assert(shift >= 64 && shift < 128);
```
which seems entirely jusstified because it's called from here:
```
auto umul192_upper64_modified(uint64_t pow10_hi, uint64_t pow10_lo,
                              uint64_t scaled_sig) noexcept -> uint64_t {
  uint128_t result =
      umul128(pow10_hi, scaled_sig) + (umul128(pow10_lo, scaled_sig) >> 64);
  uint64_t z = uint64_t(result >> 1);   <----------------------------- HERE
```
Anyway, update to follow.

vitaut:
I ran a quick test and the performance looks nearly identical to the current version with the lookup table on macOS with Apple M1.

zmij.cc
```diff
@@ -736,13 +782,13 @@ inline void write2digits(char* buffer, uint32_t value) noexcept {
...
-  write2digits(buffer + 2, bb);
-  return buffer + 4 - num_trailing_zeros[bb] -
-         (bb == 0) * num_trailing_zeros[aa];
+auto write8digits(uint32_t aa, uint32_t bb, uint32_t cc, uint32_t dd) -> uint64_t {
+  union { char b[8]; uint64_t u64; } bits;
+  write2digits(&bits.b[0], aa);
+  write2digits(&bits.b[2], bb);
+  write2digits(&bits.b[4], cc);
+  write2digits(&bits.b[6], dd);
+  return bits.u64;
```
vitaut: I'm pretty sure this is a UB in C++.

CMakeLists.txt
```diff
@@ -12,5 +12,6 @@ enable_testing()
 add_subdirectory(gtest)
 
 add_executable(zmij-test zmij-test.cc)
+target_compile_features(zmij-test PRIVATE cxx_std_20)
```
vitaut: This looks unused.

zmij.cc
```diff
@@ -710,13 +710,59 @@ inline auto digits2(size_t value) noexcept -> const char* {
...
+// MSVC makes lzcnt available always, but it's not constexpr
+inline auto lzcntl(uint64_t x) noexcept -> size_t
+{
+  return _lzcnt_u64(x);
+}
+#else
+// Both clang and gcc recognize this implementation and replace it with the
+// assembly instructions which are appropriate for the target (lzcnt or
+// bsr + zero handling).
+constexpr auto lzcntl(uint64_t x) noexcept -> size_t
```
vitaut: Let's make it not constexpr and move the `#ifdef` into a function.

zmij.cc
```diff
@@ -710,13 +710,59 @@ inline auto digits2(size_t value) noexcept -> const char* {
...
+}
+
+#ifndef _MSC_VER
+static_assert(count_trailing_nonzeros(0x30303030'30303030ull) == 0);
+static_assert(count_trailing_nonzeros(0x30303030'30303031ull) == 1);
+static_assert(count_trailing_nonzeros(0x30303030'30303039ull) == 1);
+static_assert(count_trailing_nonzeros(0x30393030'39303030ull) == 7);
+static_assert(count_trailing_nonzeros(0x31303030'30303030ull) == 8);
+static_assert(count_trailing_nonzeros(0x39303030'30303030ull) == 8);
+#endif
```
vitaut: This should go to a test and use normal gtest assertions instead of static_assert.

vitaut (changes requested):

zmij.cc
```diff
@@ -757,28 +803,29 @@ auto write_significand(char* buffer, uint64_t value) noexcept -> char* {
...
-  if (ffgghhii == 0) {
-    if (ddee != 0) return write4digits_trim_zeros(buffer, ddee);
-    return buffer - num_trailing_zeros[cc] - (cc == 0) * num_trailing_zeros[bb];
+  // Use an intermediate uint64_t to make sure that the compiler constructs
+  // the value in a register.  This way the buffer is written to memory in
+  // one go and count_trailing_nonzeros doesn't have to load from memory.
+  uint64_t buf = write8digits(bb, cc, dd, ee);
+  memcpy(buffer, &buf, 8);
+
+  if (ffgghhii == 0) [[unlikely]] {
```
vitaut, line 803: Why unlikely?

zmij.cc
```diff
@@ -710,13 +710,59 @@ inline auto digits2(size_t value) noexcept -> const char* {
...
-    "\1\0\0\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0\0\0"
-    "\1\0\0\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0\0\0";
+#ifdef _MSC_VER
+// MSVC makes lzcnt available always, but it's not constexpr
+inline auto lzcntl(uint64_t x) noexcept -> size_t
+{
+  return _lzcnt_u64(x);
+}
+#else
+// Both clang and gcc recognize this implementation and replace it with the
```
vitaut: Is MSVC not smart enough to do the same?

CMakeLists.txt
TobiSchluter: zmij-test.cc #includes zmj.cc. The latter is C++20. Without this change I run into compiler errors about using C++17 and C++20 features, but only with MSVC.

zmij.cc
TobiSchluter: Me too! I was looking for a reasonably clean form that guarantees that the value is materialized in a register, and couldn't find an option that doesn't look like
   memset((char *)some_uint64 + 2, digits(bb), 2)
which I'm fairly sure is undefined as well, no matter how I precisely address the bit range in the uint64_t. Anyway, I think I came up with something that works and is not actually undefined, I just need to confirm that the assembly looks reasonable.

CMakeLists.txt
vitaut: We only need C++17. I removed the only C++20 feature (bit_cast) earlier.

zmij.cc
TobiSchluter: I changed this to use memcpy to uint16_t variables which are then shifted and or'ed to yield the final uint64_t. The assembly ended up becoming a bit shorter as well.

zmij.cc
TobiSchluter: Done.

zmij.cc
TobiSchluter: Done

TobiSchluter:
I force-pushed a second time, because I think it's easier to review as a single commit. Sorry about the noise.

Since I didn't keep the commit messages, let me just say that there are two changes since the review:
1. I changed the string length evaluation to avoid the zero length check even if the compiler emits BSR (which gcc does by default on x64)
2. I renamed the variables that store the 8byteblocks from buf, buf2 to bits, bits2. Still not happy with the name, but there is no problem harder in programming than naming.
Edit:
3. I also accidentally removed the change which added C++20 mode for zmij-test. I still need it on MSVC.

vitaut:
Results on Apple M1 for future reference:

| Function       | Time (ns) | Speedup |
|----------------|-----------|---------|
| ostringstream  | 873.549   | 1.00x   |
| sprintf        | 734.318   | 1.19x   |
| doubleconv     | 83.789    | 10.43x  |
| to_chars       | 42.688    | 20.46x  |
| ryu            | 37.095    | 23.55x  |
| schubfach      | 24.765    | 35.27x  |
| fmt            | 22.411    | 38.98x  |
| dragonbox      | 20.654    | 42.30x  |
| zmij           | 12.006    | 72.76x  |
| zmij2          | 11.943    | 73.14x  |
| null           | 0.937     | 932.22x |

image

zmij2 is this PR.

The difference is 1% which is probably close to measurement error. However, it does make the timing more predictable and removes the lookup table so I think it's an improvement. 1-4 digit cases are handled slightly slower but I don't think it matters and the difference is tiny.

vitaut (approved):

vitaut:
Merged, thank you!

vitaut:
I also think we should handle big endian but this can be done as a follow-up.

vitaut:
BTW Paul Khuong shared his post https://pvk.ca/Blog/2017/12/22/appnexus-common-framework-its-out-also-how-to-print-integers-faster/. Looks like he used a similar technique for itoa and came to conclusion that "using registers as buffers while generating digits is a win".

TobiSchluter:
ps reading your blog post made me think how a more compact table-based approach would work. Basically, there are many unused bits in the digits2::data array. So I tried merging the length information (number of trailing zeros to be precise) into the digits array, doing something like this:
```
struct bits_and_len {
    uint16_t bits;
    size_t len;
};

inline auto digits2_len(size_t value) noexcept -> bits_and_len {
  constexpr uint16_t one = 0x4000;
  constexpr uint16_t two = 0x8000;
  constexpr uint16_t mask = 0xC000;
  constexpr size_t shift = 15;
  static const uint16_t data[] = {
    0x3030 | two, 0x3130, 0x3230, 0x3330, 0x3430, 0x3530, 0x3630, 0x3730, 0x3830, 0x3930,
    0x3031 | one, 0x3131, 0x3231, 0x3331, 0x3431, 0x3531, 0x3631, 0x3731, 0x3831, 0x3931,
    0x3032 | one, 0x3132, 0x3232, 0x3332, 0x3432, 0x3532, 0x3632, 0x3732, 0x3832, 0x3932,
    0x3033 | one, 0x3133, 0x3233, 0x3333, 0x3433, 0x3533, 0x3633, 0x3733, 0x3833, 0x3933,
    0x3034 | one, 0x3134, 0x3234, 0x3334, 0x3434, 0x3534, 0x3634, 0x3734, 0x3834, 0x3934,
    0x3035 | one, 0x3135, 0x3235, 0x3335, 0x3435, 0x3535, 0x3635, 0x3735, 0x3835, 0x3935,
    0x3036 | one, 0x3136, 0x3236, 0x3336, 0x3436, 0x3536, 0x3636, 0x3736, 0x3836, 0x3936,
    0x3037 | one, 0x3137, 0x3237, 0x3337, 0x3437, 0x3537, 0x3637, 0x3737, 0x3837, 0x3937,
    0x3038 | one, 0x3138, 0x3238, 0x3338, 0x3438, 0x3538, 0x3638, 0x3738, 0x3838, 0x3938,
    0x3039 | one, 0x3139, 0x3239, 0x3339, 0x3439, 0x3539, 0x3639, 0x3739, 0x3839, 0x3939,
  };

  return { .bits = uint16_t(data[value] & ~mask), .len = unsigned(data[value]) >> shift};
}
```

Combining this with your original approach had no impact on performance, but somehow it generates significantly more assembly. At a quick perusal I can't tell why, it doesn't seem to be the obvious suspect: no excessive bit operations in sight.

CMakeLists.txt
TobiSchluter: well sure, then it's C++17, I missed that change.

zmij.cc
TobiSchluter: My bad, I had updated the comment to say
```
// Unlike MSVC, clang and gcc recognize this implementation
```
but I should have said so explicitly over here.

Here's a godbolt to that effect
https://godbolt.org/z/Waxj54v7W

It's also of note that as of now gcc generates the zero-check even though it uses the lzcnt instruction. I'm not sure what the circumstances are that make it do that, maybe it's dependent on -march settings.

TobiSchluter:
Thanks a bunch, it was fun to dig into this.

(Sorry about the weird order of comments, github doesn't tell me if I make a comment on an outdated page ... will keep in mind to hit Ctrl+R before posting comments from now on.)

As for big-endian, I would have done it if I had a way to test aside from calling BSWAP on entering and exiting write8digits ... actually, wouldn't big-endian reverse the order in the digits2::data array? I'd need to read up on that

zmij.cc
TobiSchluter: There are 10^8 possible combinations for the trailing 8 digits but only one of those combinations takes that path.

That is of course an approximate statement given that not every decimal is representable as binary, but I assume it's close enough.

The actual distribution is probably also skewed by user input which will tend shorter. But the case where the user prints their manual input is probably not the most performance-critical?

vitaut:
> wouldn't big-endian reverse the order in the digits2::data array?

I don't think so, it's elements are just `char`s so there is nothing to reverse.

zmij.cc
vitaut: I'm not sure we can make such assumptions about the distribution of user data. It is very much possible that all of it is rounded or something like that. In any case I don't think it has significant effect so I just removed the attribute for now.

TobiSchluter:
> BTW Paul Khuong shared his post https://pvk.ca/Blog/2017/12/22/appnexus-common-framework-its-out-also-how-to-print-integers-faster/. Looks like he used a similar technique for itoa and came to conclusion that "using registers as buffers while generating digits is a win".

Interesting! I was thinking that this could be done, but didn't think it through myself. I expanded this approach to eight simultaneous digits, and it passes the godbolt test. Still need to benchmark it and get rid of some repeated masking and unmasking of 0x30,

xjb714:
@TobiSchluter In fact, on a little-endian machine, we can use the following method to calculate the BCD code of integers within the range [0, 1e8). x is the input.
```
        u64 aabb_ccdd_merge = (x << 32) - ((10000ull<<32) - 1) * ((x * 109951163) >> 40);
        u64 aa_bb_cc_dd_merge = (aabb_ccdd_merge << 16) - ((100ull<<16) - 1) * (((aabb_ccdd_merge * 10486) >> 20) & ((0x7FULL << 32) | 0x7FULL));
        u64 aabbccdd_BCD = (aa_bb_cc_dd_merge << 8) - ((10ull<<8) - 1) * (((aa_bb_cc_dd_merge * 103) >> 10) & ((0xFULL << 48) | (0xFULL << 32) | (0xFULL << 16) | 0xFULL));
```

xjb714:
On a little-endian machine, the BCD code y of x, which ranges from 0 to 99, can be calculated as follows. On the big-endian machine, y should be equal to $a  \cdot  256 + b = x + 254 \cdot (x/10)$.

$$
\begin{aligned}
a \cdot 10+b &= x\\
a+b \cdot 256 &= y = a + (x - a \cdot 10)\cdot 256 = x \cdot 256 - (10 \cdot 256 - 1)  \cdot  (x/10)
\end{aligned}
$$

xjb714:
$$
\begin{aligned}
x &= abcdefgh\\
y &= a + (b<<8) + (c<<16) + (d<<24) + (e<<32) + (f<<40) + (g<<48) + (h<<56)\\
y &= a + (b<<8) + (c<<16) + (d<<24) + ((e + (f<<8) + (g<<16) + (h<<24))<<32) \\ 
y &= a + (b<<8) + ((c + (d<<8))<<16) + ((e + (f<<8) + ((g + (h<<8)) << 16))<<32)\\
\end{aligned}
$$

`u64 aabb_ccdd_merge = (x << 32) - ((10000ull<<32) - 1) * ((x * 109951163) >> 40);`
aabb_ccdd_merge = (efgh<<32)+abcd;

`u64 aa_bb_cc_dd_merge = (aabb_ccdd_merge << 16) - ((100ull<<16) - 1) * (((aabb_ccdd_merge * 10486) >> 20) & ((0x7FULL << 32) | 0x7FULL));`
aa_bb_cc_dd_merge = (((gh<<16)+ef)<<32) + ((cd<<16)+ab);

`u64 aabbccdd_BCD = (aa_bb_cc_dd_merge << 8) - ((10ull<<8) - 1) * (((aa_bb_cc_dd_merge * 103) >> 10) & ((0xFULL << 48) | (0xFULL << 32) | (0xFULL << 16) | 0xFULL));`
aabbccdd_BCD = y = a + (b<<8) + ((c + (d<<8))<<16) + ((e + (f<<8) + ((g + (h<<8)) << 16))<<32);

xjb714:
This code is used on a big-endian CPU.

```cpp
    u64 aabb_ccdd_merge = aabbccdd + ((1ull<<32) - 10000) * ((aabbccdd * 109951163) >> 40);
    u64 aa_bb_cc_dd_merge = aabb_ccdd_merge + ((1ull<<16) - 100) * ((aabb_ccdd_merge * 10486) >> 20) & ((0x7FULL << 32) | 0x7FULL);
    u64 aabbccdd_BCD = aa_bb_cc_dd_merge + ((1ull<<8) - 10) * ((aa_bb_cc_dd_merge * 103) >> 10) & ((0xFULL << 48) | (0xFULL << 32) | (0xFULL << 16) | 0xFULL);
```

xjb714:
```cpp
#include <stdint.h>
typedef uint64_t u64;
typedef int64_t i64;

u64 bcd_little(i64 aabbccdd){
    i64 aabb_ccdd_merge = (aabbccdd << 32) + (1ull-(10000ull<<32) ) * ((aabbccdd * 109951163) >> 40);
    i64 aa_bb_cc_dd_merge = (aabb_ccdd_merge << 16) + (1-(100ull<<16) ) * (((aabb_ccdd_merge * 10486) >> 20) & ((0x7FULL << 32) | 0x7FULL));
    i64 aabbccdd_BCD = (aa_bb_cc_dd_merge << 8) + (1-(10ull<<8) ) * (((aa_bb_cc_dd_merge * 103) >> 10) & ((0xFULL << 48) | (0xFULL << 32) | (0xFULL << 16) | 0xFULL));
    return aabbccdd_BCD;
}
u64 bcd_big(i64 aabbccdd){
    u64 aabb_ccdd_merge = aabbccdd + ((1ull<<32) - 10000) * ((aabbccdd * 109951163) >> 40);
    u64 aa_bb_cc_dd_merge = aabb_ccdd_merge + ((1ull<<16) - 100) * ((aabb_ccdd_merge * 10486) >> 20) & ((0x7FULL << 32) | 0x7FULL);
    u64 aabbccdd_BCD = aa_bb_cc_dd_merge + ((1ull<<8) - 10) * ((aa_bb_cc_dd_merge * 103) >> 10) & ((0xFULL << 48) | (0xFULL << 32) | (0xFULL << 16) | 0xFULL);
    return aabbccdd_BCD;
}
```

TobiSchluter:
> @TobiSchluter In fact, on a little-endian machine, we can use the following method to calculate the BCD code of integers within the range [0, 1e8). x is the input.
> 
> ```
>         u64 aabb_ccdd_merge = (x << 32) - ((10000ull<<32) - 1) * ((x * 109951163) >> 40);
>         u64 aa_bb_cc_dd_merge = (aabb_ccdd_merge << 16) - ((100ull<<16) - 1) * (((aabb_ccdd_merge * 10486) >> 20) & ((0x7FULL << 32) | 0x7FULL));
>         u64 aabbccdd_BCD = (aa_bb_cc_dd_merge << 8) - ((10ull<<8) - 1) * (((aa_bb_cc_dd_merge * 103) >> 10) & ((0xFULL << 48) | (0xFULL << 32) | (0xFULL << 16) | 0xFULL));
> ```

That's a fun little trick to combine div and mod. I'll see if the compiler can figure it out by itself though, because that's definitely not easy to read 😆

TobiSchluter:
BCD seems to be consistently slower. I'm studying if I'm doing something dumb. The big loser appears to be not pulling the exponent from the table. Which makes sense as that calculation cannot be parallelized.

PR #8, opened 2025-12-18
fixes for x64 and arm64 msvc targets

mmozeiko:
Multiple improvements for MSVC targets:

1) better codegen for x64 using 64-bit addition & multiply intrinsics
2) better codegen for arm64 using umulh intrinsic
3) use bsr intrinsic for arm64 and non-avx2 targets instead of lzcnt. On arm64 there's no lzcnt intrinsic. And using lzcnt on x64 without bmi extension check would give wrong result as `63-value` because instruction would execute just as "bsr". The assumption is that if user enables avx2 target for msvc, then bmi is present on all such cpu's
4) fixes umul128 fallback code when doing carry in crossover sum - this calculated wrong value, because first addition was performed in 32-bit, so carry was lost
5) fixes warnings in C code when returning void function result

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -13,6 +13,10 @@
 
 #include <limits>  // std::numeric_limits
 
+#ifdef _MSC_VER
+#include <intrin.h> // lzcnt/adc/umul128/umulh
+#endif
+
 namespace {
 
 struct uint128 {
@@ -28,8 +32,14 @@ struct uint128 {
 };
 
 [[maybe_unused]] auto operator+(uint128 lhs, uint128 rhs) noexcept -> uint128 {
+#if defined(_MSC_VER) && defined(_M_AMD64)
+  uint64_t lo, hi;
+  _addcarry_u64(_addcarry_u64(0, lhs.lo, rhs.lo, &lo), lhs.hi, rhs.hi, &hi);
+  return {hi, lo};
+#else
   uint64_t lo = lhs.lo + rhs.lo;
   return {lhs.hi + rhs.hi + (lo < lhs.lo), lo};
+#endif
 }
 
 #ifdef __SIZEOF_INT128__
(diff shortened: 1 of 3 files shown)


zmij.cc
```diff
@@ -675,7 +693,7 @@ inline auto umul128(uint64_t x, uint64_t y) noexcept -> uint128_t {
   uint64_t ad = a * d;
   uint64_t bd = b * d;
 
-  uint64_t cs = uint32_t(ad) + uint32_t(bc) + (bd >> 32);  // cross sum
+  uint64_t cs = (bd >> 32) + uint32_t(ad) + uint32_t(bc);  // cross sum
```
vitaut, line 696: Why reorder?

vitaut (commented):
Thanks for the PR! LGMT, just one minor question inline.

zmij.cc
mmozeiko: That's because if you first do ` uint32_t(ad) + uint32_t(bc)` addition, then it happens only in 32-bit because both arguments are 32-bit. Only when you do next `+ (bd >> 32)` then the result of first addition is upcasted 32->64-bit and then it includes carry.

By doing `(bd >> 32) + uint32_t(ad)` first, compiler perform this addition in 64-bit, with carry. And then next `+ uint32_t(bc)` will also be in 64-bit, preserving full carry.

Alternatively you can rewrite it `uint64_t(uint32_t(ad)) + uint32_t(bc) + (bd >> 32)` if you prefer it that way better.

Without this change the test `EXPECT_EQ(dtoa(6.606854224493745e-17), "6.606854224493745e-17");` fails. Because it produces `"6.606854224493744e-17"` string.

vitaut: Good catch! We need to add a test for this 128-bit fallback logic.

PR #10, opened 2025-12-21
Make exponent handling branch-free

TobiSchluter:
A surprising speedup > 1ns, i.e. ~7% on my PC with GCC 13.3

I don't know if my version of GCC just sucks, but this was surprising. Especially taking into account that gcc recognizes the bit trick and converts it to `mov eax, edx; neg eax; cmovs edx, eax` or similar. In the graph zmij_new is the modified version, zmij is the current main (taking into account that you're working on it right now it may not be current as I type this).

image

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -999,12 +999,9 @@ void dtoa(double value, char* buffer) noexcept {
   start[1] = '.';
 
   *buffer++ = 'e';
-  char sign = '+';
-  if (dec_exp < 0) {
-    sign = '-';
-    dec_exp = -dec_exp;
-  }
-  *buffer++ = sign;
+  *buffer++ = '+' + (dec_exp < 0) * ('-' - '+');
+  int mask = dec_exp >> 31;
+  dec_exp = ((dec_exp + mask) ^ mask); // absolute value
   auto [a, bb] = divmod100(uint32_t(dec_exp));
   *buffer = char('0' + a);
   buffer += dec_exp >= 100;


zmij.cc
```diff
@@ -999,12 +999,9 @@ void dtoa(double value, char* buffer) noexcept {
   start[1] = '.';
 
   *buffer++ = 'e';
-  char sign = '+';
-  if (dec_exp < 0) {
-    sign = '-';
-    dec_exp = -dec_exp;
-  }
-  *buffer++ = sign;
+  *buffer++ = '+' + (dec_exp < 0) * ('-' - '+');
```
vitaut, line 1002: Haha, clever.

zmij.cc
```diff
@@ -999,12 +999,9 @@ void dtoa(double value, char* buffer) noexcept {
...
+  int mask = dec_exp >> 31;
```
vitaut, line 1003: This is non-portable though. Maybe `int(dec_exp < 0)` could work instead if it sign-extends?

TobiSchluter: Heh, I thought it would appear somewhere in the PR if I update the branch, anyway, I can create a new PR which changes this to something which is safe for both EBCDIC and the PDP-11
```
  *buffer++ = '-' + (dec_exp >= 0) * ('+' - '-');
  int mask = (dec_exp >= 0) - 1;
  dec_exp = ((dec_exp + mask) ^ mask); // absolute value
```

vitaut: > I can create a new PR which changes this to something which is safe for both EBCDIC and the PDP-11

Please do! We should respect old platforms =).

PR #20, opened 2025-12-22
Use std::endian if possible.

MarekKnapek:
Use std::endian if possible, as this might be more idiomatic in modern C++ than a reinterpret cast. Requires at least C++ 20 or can be detected by a feature test macro.


vitaut:
Thanks for the PR but since we need to support pre-C++20 anyway, I don't think it's worth pulling in `<bit>` and adding conditional compilation just for this.

PR #22, opened 2025-12-23
Use built-in lead zero counting instead of manual implementation

### LINKED ISSUE #21: Historic Clang versions?
It was new to me that Clang recognizes this pattern:

https://github.com/vitaut/zmij/blob/76b85b5c72ccfb2aa25ab3ebb0150813940589a1/zmij.cc#L765-L767

It indeed does. but starting Clang 19: https://godbolt.org/z/446xzn4Kf

Care about older Clang versions?

AlexGuteniev:
This makes the optimization reliable on older Clang.

Resolves #21

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -769,7 +769,7 @@ inline auto countl_zero(uint64_t x) noexcept -> int {
   // target (lzcnt or bsr + zero handling).
   int n = 64;
   for (; x > 0; x >>= 1) --n;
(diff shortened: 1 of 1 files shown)


vitaut (changes requested):
Thanks for the PR! LGMT but let's keep the generic implementation as a fallback and wrap the builtin call in `__has_builtin`.

PR #23, opened 2025-12-23
Complete the fallback for `bswap64`

AlexGuteniev:
The fallback implementation emits `bswap` instruction too for optimized builds in all major x86-64 compilers


vitaut:
Thank you!

PR #24, opened 2025-12-23
don't include `arm_neon.h` in MSVC when not compiling for ARM64

DanielaE:
Otherwise, without the guard, compilation will fail on non-ARM targets with error `C1189: #error:  This header is specific to ARM64 targets`

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -16,8 +16,10 @@
 #include <limits>       // std::numeric_limits
 #include <type_traits>  // std::conditional_t
 
-#if __has_include(<arm_neon.h>)
-#  include <arm_neon.h>
+#if !defined(_MSC_VER) || defined(_M_ARM64) || defined(_M_ARM64EC)
+#  if __has_include(<arm_neon.h>)
+#    include <arm_neon.h>
+#  endif
 #endif
 
 #ifdef _MSC_VER


vitaut (commented):
Thanks for the PR! I wonder if we should make it conditional on `__ARM_NEON__` instead of all these checks? Will it work for MSVC?

AlexGuteniev:
> I wonder if we should make it conditional on `__ARM_NEON__` instead of all these checks? Will it work for MSVC?

I'm guessing it will work in a sense MSVC won't attempt to use this even when on Arm64, where is always assumes Neon availability: https://godbolt.org/z/7qaqsjEG8

Which may or may not be your intention. The code guarded by `__ARM_NEON__` macro check doesn't look to be MSVC-friendly anyway, with designated initializer and asm statement.

vitaut:
Good point. We should definitely guard by `__ARM_NEON__` until someone makes the code compatible with MSVC.

sunmy2019:
Similar issues also happen when building on x86 linux with clang-20/clang-21.

https://godbolt.org/z/EdEWzz4cY
```
In file included from ./zmij.cc:20:
/usr/lib/llvm-20/lib/clang/20/include/arm_neon.h:28:2: error: "NEON intrinsics not available with the soft-float ABI. Please use -mfloat-abi=softfp or -mfloat-abi=hard"
   28 | #error "NEON intrinsics not available with the soft-float ABI. Please use -mfloat-abi=softfp or -mfloat-abi=hard"
      |  ^
1 error generated.
```

P.S. This PR does not fix this.

AlexGuteniev:
> until someone makes the code compatible with MSVC.

I can make it compiling.

But for running I'm unlikely to buy a Windows Arm64 device or Winows Arm64 cloud time. So my changes to compile it won't be very useful anyway, I guess.

sunmy2019:
to include llvm's <arm_neon.h>, we should check

```
// supports clang >= 11.0
# if (  defined(__arm__) || defined(__aarch64__) || defined(__arm64ec__)  ) && \
( defined(__ARM_FP) )
# endif
```
valid since https://github.com/llvm/llvm-project/commit/1b090db0df47f3ebf6acab0316180267e6b96f43

If we want to support older clang versions,
```
// supports 3.5.0 <= clang < 11.0
# if (  defined(__arm__) || defined(__aarch64__) || defined(__arm64ec__)  ) && \
( defined(__ARM_NEON) )
# endif
```

valid since https://github.com/llvm/llvm-project/commit/dee4ab08ba956efd76aa10da46510dcddecceacf

vitaut:
I replaced `__has_include` with the `__ARM_NEON` check. A PR to make tweak it to enable the optimization in more cases would be welcome.

PR #27, opened 2025-12-24
Trim leading zero from floats

dtolnay:
Before: `0.9061488e+16`
After: `9.061488e+15`

There may be a more efficient way to accomplish this.


vitaut:
Thank you!

PR #28, opened 2025-12-24
Replace memcpy with memmove for safety

AlexGuteniev:
I think if there are more digits than zeros, the ranges do overlap, so `memcpy` is not safe.

If this can be proven that the ranges do not overlap, a comment should be made. It is possible to use `memcpy` for non-overlapping parts of the same array safely.

Other occurrences of `memcpy` look safe, as they have distinct arrays as source and target.

Note that as the length is variable, no aggressive `memcpy` optimization were possible here anyway.

In MSVC, nothing bad will happen in this particular case, but formally it is still UB, and Address Sanitizer will alert.


vitaut:
Good catch, thanks!

PR #31, opened 2025-12-25
Fix 32-bit MSVC build

AlexGuteniev:
I see the other mentions of `defined(_M_AMD64)`, so I expect the support for 32-bit x86 is intended. 
<sub>I personally prefer to use `_M_X64` macro to dislike Intel and AMD equally.</sub>

MSVC intrinsics are usually silly and  map to exact instructions, so there's no `_BitScanReverse64` on 32 bit. There's also no `__lzcnt64`, but probably implementing also the same with two `__lzcnt` verstion is too much code for supporting 32-bit x86.

I don't know if the way I've written it is the most efficient; it is just the most obvious.

This time I actually built and ran test locally rather than editing on GitHub directly, all passed.

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -765,7 +765,7 @@ inline auto clz(uint64_t x) noexcept -> int {
 #elif defined(_MSC_VER) && defined(__AVX2__) && defined(_M_AMD64)
   // Use lzcnt only on AVX2-capable CPUs that have this BMI instruction.
   return __lzcnt64(x);
-#elif defined(_MSC_VER) && defined(_M_AMD64)
+#elif defined(_MSC_VER) && (defined(_M_AMD64) || defined(_M_ARM64))
   unsigned long idx;
   _BitScanReverse64(&idx, x);  // Fallback to the BSR instruction.
   return 63 - idx;


zmij.cc
```diff
@@ -762,13 +762,19 @@ inline auto clz(uint64_t x) noexcept -> int {
   assert(x != 0);
 #if defined(__has_builtin) && __has_builtin(__builtin_clzll)
   return __builtin_clzll(x);
-#elif defined(_MSC_VER) && defined(__AVX2__)
+#elif defined(_MSC_VER) && defined(__AVX2__) && defined(_M_AMD64)
   // Use lzcnt only on AVX2-capable CPUs that have this BMI instruction.
   return __lzcnt64(x);
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && defined(_M_AMD64)
```
AlexGuteniev: Alternatively, we can say `&& !defined(_M_IX86)` here, but not above.

This will make `_M_ARM64` go here.
 `_BitScanReverse64` is supported on Arm64, which is the only relevant target ISA besides x64 and x86 for current MSVC.

zmij.cc
```diff
@@ -762,13 +762,19 @@ inline auto clz(uint64_t x) noexcept -> int {
   assert(x != 0);
 #if defined(__has_builtin) && __has_builtin(__builtin_clzll)
   return __builtin_clzll(x);
-#elif defined(_MSC_VER) && defined(__AVX2__)
+#elif defined(_MSC_VER) && defined(__AVX2__) && defined(_M_AMD64)
```
vitaut, line 765: Do we need to check `_M_AMD64` here in addition to `__AVX2__`?

vitaut (commented):
Thanks for the PR!

zmij.cc
AlexGuteniev: Yes. 

`__AVX2__` is defined when `/arch:AVX2` or greater is passed for both 64-bit x86 (which also defines `_M_AMD64` and `_M_X64`) and 32-bit x86 (which also defines `_M_IX86`). 

32-bit x86 does not have `__lzcnt64` (64-bit), it only has `__lzcnt` (32-bit) and `__lzcnt16` (16-bit).

PR #32, opened 2025-12-25
Return size of the resulting representation

AlexGuteniev:


vitaut (changes requested):
I think it's easier to return the pointer to the end.

AlexGuteniev:
> I think it's easier to return the pointer to the end.

Drawbacks:
 * We need to recalculate pointer to the other buffer in the small buffer fallback.
 * We can't say `__declspec(noalias)` or other compilers equivalent, which tells the compiler that the pointer isn't aliased, and allows some angry optimizations, because when you return the pointer it actually is aliased.

Still want to proceed with that?

vitaut:
Could you elaborate on both of these points?

AlexGuteniev:
> * We need to recalculate pointer to the other buffer in the small buffer fallback.

We use temporary buffers on code paths where the buffer passed is less than expected. We need to subtract pointers, and then add the difference back to other pointer, which should be done in the header, and it is unnecessarily complex.

> * We can't say `__declspec(noalias)` or other compilers equivalent, which tells the compiler that the pointer isn't aliased, and allows some angry optimizations, because when you return the pointer it actually is aliased.

Compilers often can optimize operations with data pointed by pointers, like re-order or even vectorize, if they can guarantee that no other pointers, except the pointers the compiler sees, point to the same data. This is wide topic, I'm sure you familiar with, that includes "strict aliasing rule", `-fno-strict-aliasing`, some nasty bugs, and no less nasty missed optimizations...

Now there's function attribute that tells that the function does not save pointers it takes, doesn't return it, and so on. It helps the compiler with that aliasing  analysis by telling that if pointer snuck outside in that function the compiler doesn't see, it still didn't make any new aliases. When function returns pointer, this attribute can't apply, and if it could, the function could not return that pointer anyway.

This may be premature to apply the attribute without real use case, but it might be good to remember about such possibility.

----

Additionally, I think primary use case, which uses `string::resize_and_overwrite` also asks for size and not pointer.

vitaut (approved):
I agree that the public-facing API should have size but I was thinking that internally it might be easier to pass the pointer around. However, you convinced me that we can use size internally too.

vitaut:
Merged, thank you!

vitaut:
As it turned out we do need to return a pointer internally to avoid a small but measurable regression.

AlexGuteniev:
This also fixed the bug of negative zero size.

PR #33, opened 2025-12-25
Support more concurrency in verify program

AlexGuteniev:
The verify program assumes that hardware concurrency divides evenly on number of floats.
This effectively assumes that hardware concurrency is a power of two.

Unfortunately, this assumption is quite often not valid these days. Many CPUs produced by Intel have non power of two number of cores, like six cores. After starting to have P cores and E cores, things got wilder. For example, one popular CPU has 6 P cores with simultaneous multithreading, 8 E cores without SMT, and 2 very E cores, making up total hardware concurrency equal to 22.

We have to accept this as _the new normal_.

### DIFF
--- a/test/verify.cc
+++ b/test/verify.cc
@@ -6,7 +6,6 @@
 #include <stdio.h>   // printf
 #include <string.h>  // memcpy
 
-#include <bitset>
 #include <limits>
 #include <thread>
 #include <vector>


test/verify.cc
```diff
@@ -6,6 +6,8 @@
 #include <stdio.h>   // printf
 #include <string.h>  // memcpy
 
+#include <bitset>
```
vitaut: This looks unused.

AlexGuteniev: Yes, a stray one, I wanted  to reinvent `bit_floor` in C++17, but decided that using an odd concurrency value as it is would be easier.

vitaut:
Merged, thanks!

vitaut:
Now I can run verification on my devserver with weird number of cores where it completes in ~2 seconds.

PR #34, opened 2025-12-25
Fix MSVC warnings

AlexGuteniev:


vitaut:
Merged, thanks!

AlexGuteniev:
There's another warning added after I've started the PR, but I can't fix it as I'm not sure what to do, so perhaps it has to stay.

Here
https://github.com/vitaut/zmij/blob/6f6265936f9ce98d4e9e364b31e456fe0fc0e04c/test/dragonbox/dragonbox_to_chars.h#L82-L87

`additional_shift_amount` is zero, and in untaken ternary operand the shift is exactly by bit width, so it looks like UB, but it is never executed, so no UB actually. I.e. it is a false warning.

vitaut:
That's just a test, we can ignore it (or suppress).

AlexGuteniev:
> `additional_shift_amount` is zero, and in untaken ternary operand the shift is exactly by bit width, so it looks like UB, but it is never executed, so no UB actually. I.e. it is a false warning.

[Reported upstream](https://developercommunity.visualstudio.com/t/Undefined-behavior-is-falsely-reported-f/11021698)

AlexGuteniev:
Curiously enough, when building the dragonbox repo, I don't see any warnings

PR #36, opened 2025-12-27
Avoid subtracting unrelated pointers for small buffer

AlexGuteniev:
Subtract exactly the pointer mentioned in the same line to avoid undefined behavior, and calculate the representation size correctly for small buffer code path. 

Exercise no buffer case in the test with size usage.
Also use the size for normal buffer tests too.


vitaut:
Thanks for the fix!

PR #38, opened 2025-12-27
Fix compilation after recent changes

AlexGuteniev:
- Lambda cannot capture constexpr variable not listed in capture list, but it can capture such static constxpr variable
 - CMake doesn't want to apply C++17 for verify-double automagically

### DIFF
--- a/test/verify-double.cc
+++ b/test/verify-double.cc
@@ -35,33 +35,15 @@ const uint64_t pow10[] = {
     10000000000000000,
 };
 
-auto verify(uint64_t bits) -> bool {
-  double value = 0;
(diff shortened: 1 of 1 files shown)


AlexGuteniev:
> * CMake doesn't want to apply C++17 for verify-double automagically

Aha, the reason is is needed for verify-double but not verify-float is this:
https://github.com/vitaut/zmij/blob/88f626221258580662c69a367006d421e76ee5bb/test/verify-double.cc#L13

It fails to compile `zmij.cc`, not the `verify-double.cc` itself.

AlexGuteniev:
Room of brooms on lab island!

vitaut:
Thank you!

PR #41, opened 2025-12-28
Use SSE 4.1 for significand bcd

TobiSchluter:
This is the SIMD code for writing the significand using Xiang's BCD sequence (I can't tag him here, otherwise I would).

I did a few small things surrounding it:
- changed preprocessor logic to easier decide between different SIMD variants, this involved switching the order in the function. I chose to place the different SIMD variants in alphabetical order
- added zero checks to clz. The mask can be zero, and while it may be wasteful to use a 64bit counting instruction for a 16 bit value, it seemed more economical to reuse the 64bit function. Benchmarking showed no difference in performance. The compiler can prove that the null checks are not necessary in the existing code paths so this comes at no price
- the actual algorithm is Xiang's. I tried a few variants, but none of them performed measurably better.

Performance on my laptop is identical to the main version, on my office PC (and AMD Ryzen) it was ~10% faster when I tested before christmas.  I won't have access to that PC into the next year, so I can't confirm that I didn't mess it up since testing it on that machine, but the only change since then is the use of the 64bit bit-counting function.

I ran all tests on Windows and Linux. Benchmarks only on Linux (WSL).

As for the code which uses AVX2 256 bit registers to do the BCD conversion of the exponent at the same time as that of the significand, I plan to submit that as a separate patch. I learnt a lot about throttling and how it affects my laptop, and I explored some tricks to avoid it in benchmarking, but to no avail.  Given that this throttling can hit hard and also other applications running on the same CPU core, it would probably be an optional code path that is not used if the user expects throttling to negatively affect system performance. This should not be a concern with newer Intel architecture, especially for desktop PCs.

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -16,10 +16,6 @@
 #include <limits>       // std::numeric_limits
 #include <type_traits>  // std::conditional_t
 
-#ifdef __ARM_NEON
-#  include <arm_neon.h>
-#endif
-
 #ifdef _MSC_VER
 #  include <intrin.h>  // __lzcnt64/_umul128/__umulh
 #endif
@@ -30,6 +26,16 @@
 
 #ifndef ZMIJ_USE_SIMD
 #  define ZMIJ_USE_SIMD 1
+#endif
+
+#if ZMIJ_USE_SIMD
+#  ifdef __ARM_NEON
+#    include <arm_neon.h>
+#    define ZMIJ_USE_SIMD_NEON 1
+#  elifdef __SSE4_1__
+#    include <smmintrin.h>  // SSE4.1
+#    define ZMIJ_USE_SIMD_SSE4_1 1
+#  endif
 #endif
 
 #ifdef __has_builtin
@@ -884,22 +890,7 @@ constexpr uint64_t zeros = 0x30303030'30303030u;  // 0x30 == '0'
 // Writes a significand consisting of up to 17 decimal digits (16-17 for
 // normals) and removes trailing zeros.
 auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
-#if !defined(__ARM_NEON) || !ZMIJ_USE_SIMD
-  char* start = buffer;
-  // Each digit is denoted by a letter so value is abbccddeeffgghhii.
-  uint32_t abbccddee = uint32_t(value / 100'000'000);
-  uint32_t ffgghhii = uint32_t(value % 100'000'000);
-  buffer = write_if_nonzero(buffer, abbccddee / 100'000'000);
-  uint64_t bcd = to_bcd8(abbccddee % 100'000'000);
-  write8(buffer, bcd | zeros);
-  if (ffgghhii == 0) {
-    buffer += count_trailing_nonzeros(bcd);
-    return buffer - int(buffer - start == 1);
-  }
-  bcd = to_bcd8(ffgghhii);
-  write8(buffer + 8, bcd | zeros);
-  return buffer + 8 + count_trailing_nonzeros(bcd);
-#else   // __ARM_NEON
+#if defined(ZMIJ_USE_SIMD_NEON)
   // An optimized version for NEON by Dougall Johnson.
   struct to_string_constants {
     uint64_t mul_const;
@@ -965,7 +956,51 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
 
   buffer += 16 - (clz(~zeroes) >> 2);
   return buffer - int(buffer - start == 1);
-#endif  // __ARM_NEON
+#elif defined(ZMIJ_USE_SIMD_SSE4_1)
+  uint32_t abbccddee = uint32_t(value / 100'000'000);
+  uint32_t ffgghhii = uint32_t(value % 100'000'000);
+  uint32_t a = abbccddee / 100'000'000;
+  uint32_t bbccddee = abbccddee % 100'000'000;
+
+  buffer = write_if_nonzero(buffer, a);
+  // This BCD sequence is by Xiang JunBo.  It works the same as the one in to_bc8 but
+  // the masking can be avoided by using vector entries of the right size, and in the
+  // last step a shift operation is avoided by increasing the shift to 32 bits and
+  // then using ...mulhi... to avoid the shift.
+  __m128i x = _mm_set_epi64x(ffgghhii, bbccddee);
+  __m128i y = _mm_add_epi64(x, _mm_mul_epu32(_mm_set1_epi64x((1ull << 32) - 10000), _mm_srli_epi64(_mm_mul_epu32(x, _mm_set1_epi64x(109951163)), 40)));
+  __m128i z = _mm_add_epi64(y, _mm_mullo_epi32(_mm_set1_epi32((1 << 16) - 100), _mm_srli_epi32(_mm_mulhi_epu16(y, _mm_set1_epi16(5243)), 3)));
+  __m128i big_endian_bcd = _mm_add_epi64(z, _mm_mullo_epi16(_mm_set1_epi16((1 << 8) - 10), _mm_mulhi_epu16(z, _mm_set1_epi16(6554))));
+  __m128i bcd = _mm_shuffle_epi8(big_endian_bcd, _mm_set_epi8(8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2, 3, 4, 5, 6, 7));
+    
+  // convert to ascii
+  const __m128i ascii0 = _mm_set1_epi8('0');
+  auto digits = _mm_add_epi8(bcd, ascii0);
+ 
+  // determine number of leading zeros
+  uint16_t mask = ~_mm_movemask_epi8(_mm_cmpeq_epi8(bcd, _mm_setzero_si128()));
+  auto len = 64 - clz(mask); 
+
+  // and save result
+  _mm_storeu_si128((__m128i*)buffer, digits);
+
+  return buffer + len;
+#else   // !defined(ZMIJ_USE_SIMD_NEON) && !defined(ZMIJ_USE_SIMD_SSE4_1)
+  char* start = buffer;
+  // Each digit is denoted by a letter so value is abbccddeeffgghhii.
+  uint32_t abbccddee = uint32_t(value / 100'000'000);
+  uint32_t ffgghhii = uint32_t(value % 100'000'000);
+  buffer = write_if_nonzero(buffer, abbccddee / 100'000'000);
+  uint64_t bcd = to_bcd8(abbccddee % 100'000'000);
+  write8(buffer, bcd | zeros);
+  if (ffgghhii == 0) {
+    buffer += count_trailing_nonzeros(bcd);
+    return buffer - int(buffer - start == 1);
+  }
+  bcd = to_bcd8(ffgghhii);
+  write8(buffer + 8, bcd | zeros);
+  return buffer + 8 + count_trailing_nonzeros(bcd);
+#endif
 }
 
 // Writes a significand consisting of up to 9 decimal digits (7-9 for normals)


xjb714:
@TobiSchluter 
```uint16_t mask = ~_mm_movemask_epi8(_mm_cmpeq_epi8(bcd, _mm_setzero_si128()));```
can be replaced by
```int mask = _mm_movemask_epi8(_mm_cmpgt_epi8(bcd, _mm_setzero_si128()));```

**Comment by xjb714** [comment]
@TobiSchluter  ```int len = 32 - clz32(mask)``` or ```int len = 64 - clz64((u64)mask)```

File: zmij.cc [hunk 1]
```diff
@@ -964,7 +956,51 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
...
+  buffer = write_if_nonzero(buffer, a);
+  // This BCD sequence is by Xiang JunBo.  It works the same as the one in to_bc8 but
+  // the masking can be avoided by using vector entries of the right size, and in the
+  // last step a shift operation is avoided by increasing the shift to 32 bits and
+  // then using ...mulhi... to avoid the shift.
+  __m128i x = _mm_set_epi64x(ffgghhii, bbccddee);
+  __m128i y = _mm_add_epi64(x, _mm_mul_epu32(_mm_set1_epi64x((1ull << 32) - 10000), _mm_srli_epi64(_mm_mul_epu32(x, _mm_set1_epi64x(109951163)), 40)));
+  __m128i z = _mm_add_epi64(y, _mm_mullo_epi32(_mm_set1_epi32((1 << 16) - 100), _mm_srli_epi32(_mm_mulhi_epu16(y, _mm_set1_epi16(5243)), 3)));
+  __m128i big_endian_bcd = _mm_add_epi64(z, _mm_mullo_epi16(_mm_set1_epi16((1 << 8) - 10), _mm_mulhi_epu16(z, _mm_set1_epi16(6554))));
+  __m128i bcd = _mm_shuffle_epi8(big_endian_bcd, _mm_set_epi8(8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2, 3, 4, 5, 6, 7));
```
AlexGuteniev, line 974: This can be `_mm_shuffle_epi32` and thus have more compact encoding? Or am I missing anything?

AlexGuteniev: Ah, it is BSWAP

xjb714:
@TobiSchluter Try to calculate len using this method. Perhaps it will be faster.

zmij.cc
xjb714: @AlexGuteniev Yes，it is byteswap. The conversion from big-endian format to little-endian format.

xjb714:
@TobiSchluter 
Another way of using SSE4.1:
```cpp
    uint64_t abcdefgh = m / 100000000;
    uint64_t ijklmnop = m - abcdefgh * 100000000;
    __m128i x = _mm_set_epi64x(ijklmnop, abcdefgh);
    __m128i x_div_10000 = _mm_srli_epi64(_mm_mul_epu32(x, _mm_set1_epi32(0xd1b71759)), 45);
    __m128i x_mod_10000 = _mm_sub_epi32(x, _mm_mul_epu32(x_div_10000, _mm_set1_epi32(10000)));
    __m128i y = _mm_or_si128(x_div_10000, _mm_slli_epi64(x_mod_10000, 32));
#if defined(__SSE4_1__)
    __m128i z = _mm_sub_epi32(_mm_slli_epi32(y, 16), _mm_mullo_epi32(_mm_set1_epi32((100 << 16) - 1), _mm_srli_epi32(_mm_mulhi_epi16(y, _mm_set1_epi32(10486)), 4)));
#else
    __m128i y_div_100 = _mm_srli_epi16(_mm_mulhi_epu16(y, _mm_set1_epi16(0x147b)), 3);
    __m128i y_mod_100 = _mm_sub_epi16(y, _mm_mullo_epi16(y_div_100, _mm_set1_epi16(100)));
    __m128i z = _mm_or_si128(y_div_100, _mm_slli_epi32(y_mod_100, 16));
#endif
    __m128i z_div_10 = _mm_mulhi_epu16(z, _mm_set1_epi16(0x199a));
    __m128i little_endian_bcd = _mm_sub_epi16(_mm_slli_epi16(z, 8), _mm_mullo_epi16(_mm_set1_epi16(2559), z_div_10));
```
I'm not sure which method is faster. This method does not require byteswap.

xjb714:
@TobiSchluter SSE4.1 requires `-march=native`. If the SSE4.1 instruction set is not enabled, the above method will fallback to SSE2.

TobiSchluter:
> @TobiSchluter SSE4.1 requires `-march=native`. If the SSE4.1 instruction set is not enabled, the above method will fallback to SSE2.

Right, I refrained from adding the additional instructions before I could properly benchmark it, but on my laptop the SSE2 path doesn't pessimize things. It's a bit ridiculous that almost twenty year old technology is not enabled by default.

TobiSchluter:
> @TobiSchluter `int len = 32 - clz32(mask)` or `int len = 64 - clz64((u64)mask)`

I observed no measurable difference between these. Using the 16bit instruction is measurably slower (on my office PC)

TobiSchluter:
> @TobiSchluter `uint16_t mask = ~_mm_movemask_epi8(_mm_cmpeq_epi8(bcd, _mm_setzero_si128()));` can be replaced by `int mask = _mm_movemask_epi8(_mm_cmpgt_epi8(bcd, _mm_setzero_si128()));`

That's great(er). Saves only one NEG instruction, but it definitely is easier to read. I had only searched the listings for a not-equal instruction 🤗

zmij.cc
```diff
@@ -783,22 +789,23 @@ inline auto is_big_endian() noexcept -> bool {
 }
 
 inline auto clz(uint64_t x) noexcept -> int {
-  assert(x != 0);
-#if ZMIJ_HAS_BUILTIN(__builtin_clzll)
```
sunmy2019: Why removing this branch?
Most x86 linux users with older (like gcc <14 and clang <19) compilers with default args does not have `__builtin_ia32_lzcnt_u64` or `__builtin_clzg`.

Just use `if (x == 0) return 64;` if you want to handle zero input.

zmij.cc
```diff
@@ -783,22 +789,23 @@ inline auto is_big_endian() noexcept -> bool {
...
-  return __builtin_clzll(x);
+  #if ZMIJ_HAS_BUILTIN(__builtin_ia32_lzcnt_u64)
+  return __builtin_ia32_lzcnt_u64(x);
```
sunmy2019, line 793: `__builtin_ia32_lzcnt_u64` will silently fail on zero if it runs on CPUs not supporting lzcnt.

So you should also guard the zero.

>  It should be noted that on processors that do not support LZCNT, the instruction byte encoding is executed as BSR.

https://www.felixcloutier.com/x86/lzcnt
See also,
https://stackoverflow.com/a/47246268

zmij.cc
```diff
@@ -783,22 +789,23 @@ inline auto is_big_endian() noexcept -> bool {
...
+#elif ZMIJ_HAS_BUILTIN(__builtin_clzg)
+  return __builtin_clzg(x, 64);
```
sunmy2019, line 795: By the way, `__builtin_clzg(x, 64)` and `__lzcnt64(x)` works well on x == 0.

zmij.cc
TobiSchluter: @AlexGuteniev I actually played around with both orders in the vector (high part first or low part first), but unsurprisingly it doesn't make a difference. So I stuck with Xiang's choice.

zmij.cc
TobiSchluter: Is this built-in available if the user doesn't compile with `-mlzcnt`?

zmij.cc
TobiSchluter: Ugh, I wasn't aware that the function is a new addition. I picked the GCC 9 manual at random and it indeed isn't there. Ok, will adapt. Thanks!

zmij.cc
TobiSchluter: I noticed that `__lzcnt64` is available in gcc, but it appears to be undocumented. That's why I settled in the documented function with the long name.

TobiSchluter: Ah, that's an Intel intrinsic. Got it 👍

zmij.cc
sunmy2019: > Is this built-in available if the user doesn't compile with -mlzcnt?

No.

It's pretty safe if users know what they are doing.

But users can make mistakes and the program just silently fail.

Also sometimes a pre-built binary is shipped to the user and the user knows nothing about the compile flags.

zmij.cc
sunmy2019: > I noticed that `__lzcnt64` is available in gcc, but it appears to be undocumented. That's why I settled in the documented function with the long name.

Opps, I mean the MSVC's __lzcnt64 is safe as stated in its document. 

The gcc's __lzcnt64 is just a wrapper of the `__builtin_...`.

zmij.cc
sunmy2019: Looking through the documents again, I found that __lzcnt64 in MSVC shares the same problem.

https://learn.microsoft.com/en-us/cpp/intrinsics/lzcnt16-lzcnt-lzcnt64?view=msvc-170

sunmy2019: I found a similar issue and a fix by @AlexGuteniev 

https://github.com/microsoft/STL/issues/1103

So I think if you really need to handle zero here, the best thing you can do is to add explicit zero checks for `__builtin_clzll` and `__lzcnt` and `__builtin_ia32_lzcnt_u64`.

AlexGuteniev: I'm not sure if the same problem applies here. In MSVC STL there was expectation that lzcnt will work on any CPUs, but here we check for AVX2 macro (which is wlso done in the current MSVC STL)

In any case, I think clz should stay with just assertion, to optimize call sites that can handle zero, and here another fuction is needed that would check for zero, and it can call clz afterwards

TobiSchluter: I don't think there is an AVX2 CPU that doesn't support lzcnt. Windows 11 needs AVX2 to boot, so I think that's the expectation by MS. But I was scared away by MSVC not having an proper flag where the user has to opt in explicitly. The fact that it's encoded as `rep bsr` and thus falls back to an instruction that does the opposite is scary.

sunmy2019: > I think clz should stay with just assertion, to optimize call sites that can handle zero, and here another fuction is needed that would check for zero, and it can call clz afterwards

I share the same idea. It's clear and clever.

sunmy2019: > The fact that it's encoded as rep bsr and thus falls back to an instruction that does the opposite is scary.

I took a deeper look. It's because x86 `rep bsr` and `lzcnt` share the same encoding `F3 0F BD`, and  only newer CPUs will decode it as the latter. What a x86!

TobiSchluter: @sunmy2019 just what i said :D

I think the right thing is to add a sentinel bit as I did previously, because I found that gcc 13.3 sometimes keeps the null checks as jumps even with lzcnt (!). I can't reproduce it right now, but I also remember seeing the compiler seeing `(x != 0) * __builtin_clz(x)` and thinking that since the second factor is undefined if `x == 0` it can assume that the whole product is undefined, and thus transforming it to a simple unchecked `__builtin_clz(x)`

vitaut: +1 to adding a sentinel if doesn't regress perf too much.

zmij.cc
```diff
@@ -783,22 +789,23 @@ inline auto is_big_endian() noexcept -> bool {
...
 #elif defined(_MSC_VER) && defined(__AVX2__) && defined(_M_AMD64)
   // Use lzcnt only on AVX2-capable CPUs that have this BMI instruction.
   return __lzcnt64(x);
 #elif defined(_MSC_VER) && (defined(_M_AMD64) || defined(_M_ARM64))
   unsigned long idx;
-  _BitScanReverse64(&idx, x);  // Fallback to the BSR instruction.
-  return 63 - idx;
+  auto nonzero_p = _BitScanReverse64(&idx, x);  // Fallback to the BSR instruction.
```
vitaut, line 801: Why `auto`?

zmij.cc
```diff
@@ -32,6 +28,16 @@
 #  define ZMIJ_USE_SIMD 1
 #endif
 
+#if ZMIJ_USE_SIMD
+#  ifdef __ARM_NEON
+#    include <arm_neon.h>
+#    define ZMIJ_USE_SIMD_NEON 1
```
vitaut, line 34: I think we can define it without specific value since the latter is not used.

zmij.cc
TobiSchluter: Autof habit, I noticed that your style is to avoid autos, and will fix that. (Unless the sentinel bit works out.)

zmij.cc
TobiSchluter: BTW I filed a related missed-optimization bug against GCC. Turns out that 64 is a bad number to multiply with a boolean. https://gcc.gnu.org/PR123330

vitaut (changes requested):
Marking as changes requested since some comments still need to be addressed.

TobiSchluter:
Hi, happy new year, I've been busy with the festivities and spent all the time I had on trying to figure out how to benchmark reliably on my laptop (still an unsolved problem though I was able to make it more reproducible), and I also did some experiments with the code. I think the biggest surprise was that preloading the constants for multiplication, byte swapping, and ascii conversion into the many registers that gcc leaves unused didn't change the performance at all even if I made sure to load them only once in a whole benchmarking run.  I'm therefore suspecting that writing the xmm register and then immediately modifying the same memory for the decimal point and exponent insertion is what actually dominates the execution time. I had some comical results which support that: using a masked move which only writes the bytes before the trailing zeros made the execution time increase eightfold! Less work in more time. Well, it seems that it's a really bad idea to touch the same memory that a masked move touches.

I have some an idea how to handle this and at least do the decimal point with no overwriting of the 16 bytes that xmm registers give me, and will try this before finishing up this patch.

vitaut:
BTW you can now use this benchmark which should give much more stable results: https://github.com/vitaut/zmij/blob/main/test/benchmark.cc

TobiSchluter:
> @TobiSchluter SSE4.1 requires `-march=native`. If the SSE4.1 instruction set is not enabled, the above method will fallback to SSE2.

@xjb714 it took me a while to grasp the import of this message, only when I reviewed which instruction is introduced by which SSE version did I realize the `_mm_shuffle_epi8` used in the big-endian version is not an SSE2 instruction, but SSE3.  Your low-endian version came out slower than the big-endian version, so I played around a bit with ways of using SSE2 only, and I came up with this hybrid of both as an SSE2-only version (sorry about partly inlining the multiplication constants, this is WIP):
```
  __m128i x = _mm_set_epi64x(ffgghhii, bbccddee); // right digits first, left digits second
  __m128i y = _mm_add_epi64(x, _mm_mul_epu32(divmod10000, _mm_srli_epi64(_mm_mul_epu32(x, div10000), 40)));
    __m128i y_div_100 = _mm_srli_epi16(_mm_mulhi_epu16(y, div100), 3);
    __m128i y_mod_100 = _mm_sub_epi16(y, _mm_mullo_epi16(y_div_100, one_hundred));
    __m128i z = _mm_slli_epi32(y_mod_100, 16) | y_div_100;
    __m128i z_div_10 = _mm_mulhi_epu16(z, _mm_set1_epi16(0x199a));
    __m128i bcd_swapped = _mm_sub_epi16(_mm_slli_epi16(z, 8), _mm_mullo_epi16(_mm_set1_epi16(2559), z_div_10));
    __m128i bcd = _mm_shuffle_epi32(bcd_swapped, _MM_SHUFFLE(2, 3, 0, 1));
```
It reverses the 8 digit parts in the beginning and the 2 digit as well as single digit parts as they are calculated. The 4 digits, i.e. 32bit, parts can be shuffled even in SSE2. In an isolated benchmark (i.e. not inlined into zmij) which uses `rdtsc` to evaluate the number of cycles measure this as ~11 cycles on my office PC (AMD Ryzen) compared to ~16 cycles for the non-SIMD version and ~8 cycles for the SSE 4.2 fast path (a null function which only returns an updated `char*` takes three cycles in this measurement).

I think the only cases worth optimizing for are:
- default settings (i.e. SSE2)
- `-march=native` on a recent CPU (i.e. at least SSE4.2, AVX or AVX2 if it can be proven to be faster)

I think only the latter makes sense for MSVC, as Windows 11 needs AVX.

As always, I appreciate your input.

Edit: I should have said, your version runs at the same speed, but it's a few instructions longer. https://godbolt.org/z/oo76f3zs4
Edit: seems like Windows 11 actually needs sse4.2 + popcnt, so no free lzcnt either :(

xjb714:
@TobiSchluter Replace ```__m128i z = _mm_slli_epi32(y_mod_100, 16) | y_div_100;``` with ```__m128i z = _mm_or_si128(y_div_100, _mm_slli_epi32(y_mod_100, 16));``` so that it can be compiled under the msvc compiler.

TobiSchluter:
>I'm therefore suspecting that writing the xmm register and then immediately modifying the same memory for the decimal point and exponent insertion is what actually dominates the execution time. I had some comical results which support that: using a masked move which only writes the bytes before the trailing zeros made the execution time increase eightfold! Less work in more time. Well, it seems that it's a really bad idea to touch the same memory that a masked move touches.

So, the issue with masked moves was that benchmarking repeatedly wrote to the same memory, and that is terrible for _mm_maskmoveu_si128. Cycling through a number of different buffers speeds it up significantly, but it's still slower than just writing the whole register to memory (32 cycles vs 10 cycles with the same caveats as above)

AlexGuteniev:
> So, the issue with masked moves was that benchmarking repeatedly wrote to the same memory, and that is terrible for _mm_maskmoveu_si128

I assume this is non-temoral operation, so is designed that you'll not touch that memory afterwards.

It also needs barriers, if you measure without it, you might be measure before store completion.

It also has terrible timings on AMD: https://uops.info/table.html?search=maskmovdqu&cb_lat=on&cb_tp=on&cb_uops=on&cb_ports=on&cb_ADLP=on&cb_ADLE=on&cb_ZEN4=on&cb_measurements=on&cb_doc=on&cb_base=on&cb_sse=on

I think you should only use non-temporatl ops for optimizations with large data, not micro optimizations like dtoa, and even for that strive to use non-masked stores.

TobiSchluter:
> It also has terrible timings on AMD: https://uops.info/table.html?search=maskmovdqu&cb_lat=on&cb_tp=on&cb_uops=on&cb_ports=on&cb_ADLP=on&cb_ADLE=on&cb_ZEN4=on&cb_measurements=on&cb_doc=on&cb_base=on&cb_sse=on

Heh, that's a fun webpage. Thanks for the pointer. The masked move thing was just an aside, but somewhat interesting as I'm exploring putting together the whole string including exponent in a 256bit register, which would either require a larger buffer or a masked move.

AlexGuteniev:
> Heh, that's a fun webpage. Thanks for the pointer

In case you didn't know there's another source of info of this level, it is Agner Fog's tables. See https://www.agner.org/optimize/ book 4. Instruction tables. But I prefer https://uops.info for convenient access.

> I'm exploring putting together the whole string including exponent in a 256bit register, which would either require a larger buffer or a masked move.

Just don't try to use that one SSE2 masked store.

AVX2 masked store is fine if not on AMD. It has 32-bit granularity though. 
AVX-512 masked store should be fine anywhere where AVX-512 is functioning normally. But functioning AVX-512 is still rare enough.
The best option, except buffer expanding, should be two overlapped 128bit stores.

TobiSchluter:
Here are the timings of the patch I'm planning to submit tonight. Note that this uses `-mlzcnt`, otherwise there is an unaesthetical uptick at 2 digits (not a common case in FP I would guess). This happens because gcc doesn't optimize zero-checked `clz` very well, see my bug report https://gcc.gnu.org/PR123330
image

TobiSchluter:
Moved to new pull request #59

PR #43, opened 2025-12-29
Don't try to use hundreds in float exponent

AlexGuteniev:
The improvement is barely measurable, but I think it is still good to do on principle to avoid dead code.


vitaut:
Thank you!

PR #45, opened 2025-12-29
Fix compilation and use C++20 in verification programs

### LINKED ISSUE #44: non-static constexpr local variables captured in verify-double
Previous commit: https://github.com/vitaut/zmij/pull/38/commits/8183f5f0a407e1888683be0a8835f4dcfbe8d5dc

This happened again:
```
zmij\test\verify-double.cc(167): error C3493: 'exp_shift' cannot be implicitly captured because no default capture mode has been specified
zmij\test\verify-double.cc(168): error C3493: 'pow10_lo' cannot be implicitly captured because no default capture mode has been specified
```

The implicit capturing of such variables is rejected by MSVC.

Such behavior is an _extension_. So this behavior that can be disabled by any of the following:
 * Passing `/permissive-` to MSVC to disable all extensions
 * Using C++20 or newer to implicitly use `/permissive-` mode
 * Passing `/Zc:lambda` to disable the particular _extension_

Alternatively we can add more `static`s. Or capture these constants.

I think we should pass  `/permissive-` and remove two `static`s added previously.

AlexGuteniev:
Fix #44 in some implicit way.

Bump up C++ standard to C++20.
In MSVC it causes implicit strict language conformance enabling.
Among some other things, this mode allows MSVC to accept some valid C++ code that is not accepted by default.

Also undo some previous adaptation to not having this mode enabled.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,3 +1,5 @@
+set(ZMIJ_VERIFIER_STANDARD cxx_std_20)
+
 add_subdirectory(gtest)
 
 add_library(dragonbox dragonbox/dragonbox_to_chars.cpp)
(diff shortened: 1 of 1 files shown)


vitaut:
Thank you!

PR #47, opened 2025-12-31
make `zmij` usable at compile time, or as header-only library

DanielaE:
Generated code with compilation mode 2 (see below):
https://godbolt.org/z/T5K5srdT4     x64, trunk compilers, C++17 max. optimization
https://godbolt.org/z/dTa4vMnMT  ARM64, trunk compilers, C++17 (except gcc) max. optimization

I can see no impact on code generation.

There are two additional compilation modes besides the current split between header+implementation:

1. header-only (when `ZMIJ_HEADER_ONLY` is defined)
    this requires the introduction of external linkage for all entities, and `inline` definitions where necessary

2. enable _constant evaluation_ (when ZMIJ_ENABLE_COMPILETIME_EVALUATION is defined)
    * this necessarily implies header-only
    * the separate compilation and use of `zmij.cc` is still possible according to my tests with msvc, and generates identical binaries.
    * `constexpr` where existing functions are implicitly or explicitly `inline`
    * some inherently non-`constexpr`- capable functions get replaced with `constexpr` ones
    * keep compatibility with C++17
  
In general:
 * detect the availability of compiler features to select the best language-/library-feature to use
 * if a compiler is not decent enough to support constant evaluation, either bail ot with an error or fall back to pure runtime code and #undef macro ZMIJ_ENABLE_COMPILETIME_EVALUATION before the end of `zmij.h`

I expect comments and opinions, or possibly even push-back of this idea.  Hence, this is only a draft PR for now.

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -147,8 +147,10 @@ struct dec_fp {
 
 #  ifdef ZMIJ_HAS_IF_CONSTEVAL
 #    define ZMIJ_IF_CONSTEVAL if consteval
+#    define ZMIJ_IF_RUNTIME if !consteval
 #  elif defined(ZMIJ_HAS_TEST_COMPILETIME)
 #    define ZMIJ_IF_CONSTEVAL if (__builtin_is_constant_evaluated())
+#    define ZMIJ_IF_RUNTIME if (!__builtin_is_constant_evaluated())
 #  endif
 
 template <typename T>
@@ -194,6 +196,7 @@ ZMIJ_CONSTEVAL auto ct_memcpy(char* dst, const T* src, unsigned len) noexcept
 #  define ZMIJ_INLINE inline
 #  define ZMIJ_CONSTEXPR
 #  define ZMIJ_IF_CONSTEVAL if constexpr (false)
+#  define ZMIJ_IF_RUNTIME if constexpr (true)
 
 #  define ZMIJ_MEMCPY memcpy
 #  define ZMIJ_MEMMOVE memmove
@@ -268,10 +271,10 @@ template <typename Float> struct float_traits : std::numeric_limits<Float> {
 #ifdef ZMIJ_HAS_BITCAST
     return __builtin_bit_cast(sig_type, value);
 #else
-    uint64_t bits = 0;
-    ZMIJ_IF_CONSTEVAL {}
-    else {
+    ZMIJ_IF_RUNTIME {
+      uint64_t bits = 0;
       memcpy(&bits, &value, sizeof(value));
+      return bits;
     }
 #endif
   }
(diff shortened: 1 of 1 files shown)


AlexGuteniev:
I find the `#undef INLINE` way too malicious, and think it should follow ZMIJ_INLINE practice.
Also observe it is Polish Żmij and not some ZMEJ.

DanielaE:
Besides the unintended ZEMJ vs ZMIJ typo (my Polish isn't polished enough to immediately spot that), what do you mean with your first sentence?

AlexGuteniev:
Some user may define INLINE before Żmij, e.g. in a precompiled header and expect that to be intact. An acceptable solution is to use a library prefix for each macro, in hope there are no other Żmijs included.

example.cc
```diff
@@ -1,7 +1,46 @@
...
+static_assert(to_string(1.3421781e+08f) == "1.3421781e+08");
+static_assert(to_string(std::numeric_limits<double>::denorm_min()) == "5e-324");
+static_assert(to_string(std::numeric_limits<float>::denorm_min()) == "1e-45");
+static_assert(to_string(1e-323) == "1e-323");
+static_assert(to_string(0.) == "0");
+static_assert(to_string(-0.f) == "-0");
+static_assert(to_string(std::numeric_limits<double>::infinity()) == "inf");
+static_assert(to_string(-std::numeric_limits<float>::infinity()) == "-inf");
+static_assert(to_string(-std::numeric_limits<double>::quiet_NaN()) == "-nan");
+static_assert(to_string(std::numeric_limits<float>::quiet_NaN()) == "nan");
```
vitaut: Let's not pollute example with this.

vitaut (changes requested):
Thanks for the PR but I think all the conditional compilation adds too much clutter. In general this implementation is optimized for runtime performance and doesn't make much sense to use directly at compile time. For the latter we could use a basic Schubfach.

AlexGuteniev:
> I think all the conditional compilation adds too much clutter

Some clutter can be avoided by using the pattern like here,  i.e. sharing the fallback for consteval and non-intrinsic instead of copying it:
https://github.com/vitaut/zmij/blob/08772ee5ecabc144345535bfd1889fc9d6b21040/zmij.cc#L168-L188

I'm not sure if it can make the PR better though. 

Ultimately I agree with not supporting compile-time formatting, or having distinct implementation, instead of trying to gen the whole Żmij to compile time. The only potential advantage I see is that core language UB in compile time is a hard error, so compile time testing may be helpful in this regard. But it is still not what compile time is generally for.

DanielaE:
> Thanks for the PR but I think all the conditional compilation adds too much clutter. In general this implementation is optimized for runtime performance and doesn't make much sense to use directly at compile time. For the latter we could use a basic Schubfach.

I'm not too surprised about this opinion, and I feel to some degree similar. In particular, using basic Schubfach as the foundation of a venue towards missing `constexpr`-enabled pieces in the standard.

OTOH, staying as close as possible to the runtime implementation is the - by far - best option to *guarantee* freedom of UB (of *any* C++ source code) with the help of extensive compile time testing. AFAIK, no other measure can give you that, no sanitizer, no nothing.

Anyway, thanks for looking at my PR. It was a good excercise in what you can do towards the final goal of  'constexpr all the things', going even beyond most people's perceived limitations and expectations. You mentioned Hana elsewhere, I agree with that.

vitaut:
Closing per discussion.

PR #48, opened 2025-12-31
Fix compile-time table generation in MSVC

AlexGuteniev:
Annoyingly, the majority of MSVC intrinsics do not work in compile time.

Fortunately `__builtin_is_constant_evaluated()` works, and does it regardless of the standard mode.


vitaut:
Thank you!

PR #52, opened 2026-01-01
Reduce float_buffer_size from 17 to 16

### LINKED ISSUE #51: The buffers seem too big
I see the following limits:
https://github.com/vitaut/zmij/blob/3e248165f26fddc5eddf6c2e90360ae9f762d66c/zmij.h#L35-L36

Looks like the longest float representation is in a form `-d.dddddddde±dd` which has 15 digits, so only 16 buffer is needed, including the room for `'\0'`. 
The 25 probably should be double checked too.

dtolnay:
Size 16 is big enough to fit `-1.00000005e+15` (15 characters) plus nul terminator. 17 is unnecessary for any float representation currently produced by zmij::write.

Closes #51.

### DIFF
--- a/zmij.h
+++ b/zmij.h
@@ -33,7 +33,7 @@ auto to_decimal(double value) noexcept -> dec_fp;
 
 enum {
   double_buffer_size = 25,
-  float_buffer_size = 17,
+  float_buffer_size = 16,
 };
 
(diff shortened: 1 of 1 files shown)


vitaut:
Thank you!

PR #54, opened 2026-01-01
Add buffer overrun tests

AlexGuteniev:
Values from #51 discussion to have longest representations, so that any longer buffer write will trigger an overrun.

(Of course the overrun if present will also make the whole program behavior undefined, due to having other tests that don't expect overrun, and due to reserving just one sentinel character, so philosophically this is not a reliable test)


vitaut:
This is great, thanks!

PR #55, opened 2026-01-03
Compile Neon vectorization on Arm64 MSVC

AlexGuteniev:
I'm not sure if this works in MSVC, but at least it compiles, and seem to be minimal changes.

Changes:
1. Preprocessor `defined(_MSC_VER) && defined(_M_ARM64)` is the only condition for Neon, we assume Neon as a requirement of Windows on Arm64, and don't check for it.
2. Stuff the values into the structure with constants member initializers, rather than having designated initializer. MSVC only accepts designated initializer in C++20 or newer mode.
3. Use arrays in the structure with constants. These opaque types for vector intrinsic aren't necessarily arrays, but the code uses them as arrays, and not as vector intrinsic types.
4. Compiler barriers. MSVC doesn't have functioning `asm` statement. ~~There are `_ReadWriteBarrier()`, `_ReadBarrier()`, `_WriteBarrier()`, but these are for the whole code. For particular variables there are special load/store intrinsic that serve the same purpose as `asm` statement. Unfortunately, not for vector-sized variables.~~ no barriers for now.
5. MSVC lacks built-in 128 bit integers, used the integer class instead.

The uint128 change and the vector types change are applied unconditionally, not just for MSVC, so please check if the perf isn't ruined.

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -356,7 +356,7 @@ constexpr uint64_t zeros = 0x0101010101010101u * '0';
 // Writes a significand consisting of up to 17 decimal digits (16-17 for
 // normals) and removes trailing zeros.
 auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
-#  if !ZMIJ_USE_NEON
+#if !ZMIJ_USE_NEON
   char* start = buffer;
   // Each digit is denoted by a letter so value is abbccddeeffgghhii.
   uint32_t abbccddee = uint32_t(value / 100'000'000);
@@ -371,13 +371,13 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
   bcd = to_bcd8(ffgghhii);
   write8(buffer + 8, bcd | zeros);
   return buffer + 8 + count_trailing_nonzeros(bcd);
-#else   // ZMIJ_USE_NEON
+#else  // ZMIJ_USE_NEON
   // An optimized version for NEON by Dougall Johnson.
   struct to_string_constants {
     uint64_t mul_const = 0xabcc77118461cefd;
     uint64_t hundred_million = 100000000;
     int32_t multipliers32[4] = {0x68db8bb, -10000 + 0x10000, 0x147b000,
-                                   -100 + 0x10000};
+                                -100 + 0x10000};
     int16_t multipliers16[8] = {0xce0, -10 + 0x100};
   };
 
@@ -389,7 +389,7 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
   // Compiler barrier, or clang doesn't load from memory and generates 15 more
   // instructions
   asm("" : "+r"(c));
-#endif
+#  endif
 
   uint64_t hundred_million = c->hundred_million;
 


zmij.cc
```diff
@@ -23,7 +23,7 @@ struct dec_fp {
 #include <limits>       // std::numeric_limits
 #include <type_traits>  // std::conditional_t
 
-#ifdef __ARM_NEON
+#if defined(__ARM_NEON) || defined(_MSC_VER) && defined(_M_ARM64)
```
vitaut: Let's move this condition into a new macro, `ZMIJ_USE_NEON`.

Sir-Walrus: Maybe also add a paren, neither I [nor gcc](https://godbolt.org/z/Wrenf8e7c) like && beside ||.

AlexGuteniev: Aye, sir!

AlexGuteniev: > Maybe also add a paren, neither I [nor gcc](https://godbolt.org/z/Wrenf8e7c) like && beside ||.

The new parens culture hasn't reached the preprocessor yet: https://godbolt.org/z/o5fTKK4x1

I think we can though. @vitaut, thoughts?

vitaut: I am in favor of parentheses.

zmij.cc
```diff
@@ -347,7 +355,7 @@ constexpr uint64_t zeros = 0x0101010101010101u * '0';
 // Writes a significand consisting of up to 17 decimal digits (16-17 for
 // normals) and removes trailing zeros.
 auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
-#if !defined(__ARM_NEON) || !ZMIJ_USE_SIMD
+#  if !ZMIJ_USE_NEON || !ZMIJ_USE_SIMD
```
vitaut: `ZMIJ_USE_SIMD` check can also be moved to the definition of `ZMIJ_USE_NEON`.

zmij.cc
```diff
@@ -388,6 +393,17 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
 
   // Compiler barrier, or clang narrows the load to 32-bit and unpairs it.
   asm("" : "+r"(hundred_million));
+#  else
+  // Compiler barrier, or clang doesn't load from memory and generates 15 more
```
vitaut: Why clang?

zmij.cc
```diff
@@ -388,6 +393,17 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
...
+  // instructions
+  const to_string_constants* c;
+  __iso_volatile_store64(reinterpret_cast<volatile int64_t*>(&c),
```
vitaut: Does it actually need volatile?

AlexGuteniev: The signature of all atomic intinsics includes `volatile`, see https://learn.microsoft.com/en-us/cpp/intrinsics/arm64-intrinsics for `__iso_volatile_store64` on Arm64 in particular. It is fine do add that via a cast, `atomic_ref` works this way.

The real questions are whether this is correct translation of the original code intention, and whether we need this for MSVC. I don't know really.

zmij.cc
AlexGuteniev: I don't know if MSVC even needs this.

vitaut: I mean it's guarded on `_MSC_VER` which seems to contradict the comment. Or clang is used as a backend?

zmij.cc
vitaut: Have you tried comparing the assembly?

zmij.cc
AlexGuteniev: > I mean it's guarded on `_MSC_VER` which seems to contradict the comment

The comment is for what MSVC would have done it it was Clang

AlexGuteniev: > Or clang is used as a backend?

Do you mean clang-cl? 
It doesn't need the change proposed in this PR.

Instead it fails differently, both before and after the PR with this error:
```
C:\Project\zmij\out\build\x64-Clang-Release\lld-link : error : undefined symbol: __udivti3
  
  >>> referenced by C:\Project\zmij\test\modular-search.h:55
  
  >>>               test\CMakeFiles\double-check.dir\double-check.exe.lto.double-check.cc.obj:(unsigned __int64 __cdecl find_carried_away_doubles<7784369436827535057, 3, class `public: <auto> __cdecl `extern "C" main'::`1'::<lambda_0>::operator()(void) const'::`1'::<lambda_1>, class `public: <auto> __cdecl `extern "C" main'::`1'::<lambda_0>::operator()(void) const'::`1'::<lambda_2>>(unsigned __int64, unsigned __int64, class `public: <auto> __cdecl `extern "C" main'::`1'::<lambda_0>::operator()(void) const'::`1'::<lambda_1>, class `public: <auto> __cdecl `extern "C" main'::`1'::<lambda_0>::operator()(void) const'::`1'::<lambda_2>))
  
  >>> referenced by C:\Project\zmij\test\modular-search.h:55
  
  >>>               test\CMakeFiles\double-check.dir\double-check.exe.lto.double-check.cc.obj:(unsigned __int64 __cdecl find_carried_away_doubles<7784369436827535057, 3, class `public: <auto> __cdecl `extern "C" main'::`1'::<lambda_0>::operator()(void) const'::`1'::<lambda_1>, class `public: <auto> __cdecl `extern "C" main'::`1'::<lambda_0>::operator()(void) const'::`1'::<lambda_2>>(unsigned __int64, unsigned __int64, class `public: <auto> __cdecl `extern "C" main'::`1'::<lambda_0>::operator()(void) const'::`1'::<lambda_1>, class `public: <auto> __cdecl `extern "C" main'::`1'::<lambda_0>::operator()(void) const'::`1'::<lambda_2>))
  
  >>> referenced by C:\Project\zmij\test\modular-search.h:55
  
  >>>               test\CMakeFiles\double-check.dir\double-check.exe.lto.double-check.cc.obj:(unsigned __int64 __cdecl find_carried_away_doubles<7784369436827535057, 3, class `public: <auto> __cdecl `extern "C" main'::`1'::<lambda_0>::operator()(void) const'::`1'::<lambda_1>, class `public: <auto> __cdecl `extern "C" main'::`1'::<lambda_0>::operator()(void) const'::`1'::<lambda_2>>(unsigned __int64, unsigned __int64, class `public: <auto> __cdecl `extern "C" main'::`1'::<lambda_0>::operator()(void) const'::`1'::<lambda_1>, class `public: <auto> __cdecl `extern "C" main'::`1'::<lambda_0>::operator()(void) const'::`1'::<lambda_2>))
  
  
  >>> referenced 14 more times
```
Happens for x64 too.

Think it is llvm/llvm-project#25679

zmij.cc
AlexGuteniev: Just checked. There's certainly a difference, and there's certainly more code without barriers.

My limited understanding doesn't allow me to tell which is better.

FmOavylakAAT0l7

zmij.cc
vitaut: > The comment is for what MSVC would have done it it was Clang

I don't think it's clear from the comment. If there is no evidence that it makes things faster on MSVC then I suggest simplifying this to just 

```
const to_string_constants* c = &constants;
```

AlexGuteniev: Ok, done that.

Note that we can't just leave `asm` statements in MSVC, because GNU asm statement is non-standard, the standard asks for ` asm ( string-literal ) ;`, so MSVC can't ignore it and errors out there.

AlexGuteniev: Edited the description to reflect the approach change.

vitaut:
Merged, thanks!

PR #58, opened 2026-01-05
Fuse stores for 32-bit floats as well

AlexGuteniev:
Minor perf improvement, similar to fusing hundreds and the rest for doubles.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,8 +3,6 @@ project(zmij CXX)
 
 set(ZMIJ_STANDARD cxx_std_17)
 
-set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)
-
 add_library(zmij zmij.cc zmij.h)
 target_include_directories(zmij PUBLIC .)
 target_compile_features(zmij PRIVATE ${ZMIJ_STANDARD})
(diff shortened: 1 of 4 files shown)


AlexGuteniev:
An easy way to observe the improvement is to run `float-check` few times before and after the change, at least that's what I did.

vitaut:
Looks like there is a merge conflict.

AlexGuteniev:
Aha, as discussed elsewhere you removed `constexpr` to compile using Borland C++ 4.0

I've resolved on GitHub directly, but I hope I've done it right.

vitaut:
Running the benchmark with a quick and dirty tweak casting value to `float` I don't see a difference on M1:

Before:

```
dragonbox: 10.40ns ( 9.86ns - 10.65ns)
zmij     :  6.92ns ( 6.88ns -  6.96ns)
```

After:
```
dragonbox: 10.42ns ( 9.86ns - 10.64ns)
zmij     :  6.95ns ( 6.90ns -  7.00ns)
```

AlexGuteniev:
> I don't see a difference on M1

Can you try an x86-64?

vitaut:
I only have amd64 but can try

AlexGuteniev:
> I only have amd64

I meant exactly that

vitaut:
Numbers are not very stable but it doesn't look like there is much difference there either:

Before:
```
dragonbox: 14.56ns (13.89ns - 14.92ns)
zmij     : 11.08ns (10.77ns - 11.30ns)
```

After:
```
dragonbox: 14.55ns (13.83ns - 14.94ns)
zmij     : 11.01ns (10.67ns - 11.26ns)
```

AlexGuteniev:
Ok, then it is either a hallucination or some sudden codegen effect on my computer

PR #7, opened 2025-12-16
Use BCD coding to evaluate eight significand digits in parallel

TobiSchluter:
… instead of table lookups. This similar to what https://mastodon.social/@pkhuong@discuss.systems described, but for eight digits (and I of course avoided adding 0x30s before counting bits, as Paul also mentioned). The bcd sequence doesn't use [xjb714](https://github.com/xjb714)'s tricks that he described in https://github.com/vitaut/zmij/pull/4. The generated code when using them is one insturction shorter, but involves one more multiplication, which is the same performance-wise, but I find this much more readable.

No performance impact, code gets shorter by 13 instructions in my godbolt test 
    main branch: https://godbolt.org/z/c68jGa6vv
    this: https://godbolt.org/z/977E3371o

A first attmept at using 128bit BCD turned out much slower. I also tried various ways to get rid of the division to extract `a` (i.e. the usually zero first digit) but they all had negative performance impact.

I couldn't remove the digits2 table yet because it is also used when formatting the exponent where a similar pseudo-smd bcd trick is not available. Using the limited range of numbers (abs(exp) < 350) allows simplifying the calculation but no matter which way I phrase the direct calculation I see a fairly consistent degradation by 1ns, i.e. 5%.

PR #59, opened 2026-01-06
SSE simd support

TobiSchluter:
This patch adds SSE2 and SSE 4.1 support to `write_significand17`. These two SSE versions are interesting because SSE2 is enabled by default in most compilers and SSE4.1 allows for saving some operations. It continues a discussion started in https://github.com/vitaut/zmij/pull/7 and supercedes the previous pull request https://github.com/vitaut/zmij/pull/41.

Let me preface this by saying that this is not always a performance gain: on my laptop (CPU: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz)  the SSE version is as fast as the scalar version as long as the significand has at least 9 digits.

The world looks more beautiful on a proper machine though, so all benchmarks that follow are on an AMD Ryzen 9 9950X.

To quickly summarize what's going on here: the concept is similar to `to_bcd8, but expanded to 16 digits. We successively transform the number into bases 100'000'000, 10'000, 100, and finally 10. Where the digits are handled in paralle. Once this is done, we copy the whole string to the output and use bitcounting to determine the number of trailing zeros that we want the caller to ignore and return the appropriately advanced buffer. This uses BCD sequences derived from code that Xiang JunBo provided, and I took some inspiration from Dougall Johnson's code. It uses one of two BCD sequences, one for SSE2 (where no 32bit mullo and no complete swap is available) and an SSE4.1 version which harnesses both.

Besides the BCD sequence there is another choice made: namely the use of the `LZCNT` instruction to extract the length: the zero check required for `clz` turns out to be expensive, at least with GCC. A bug report to that effect is here https://gcc.gnu.org/PR123330. Note that the only case where there can actually be zero fed into `clz` is the case where the mantissa has 17 digits and only the first digit is set. I tried an early return in that case, and it accelerates the single-digits case, but  that is probably not a case worth optimizing for.

Lastly, you will see a 128byte large block of constants embedded in the code. Two remarks on this: calculating the constants every time costs approx. 0.3ns, so those 128bytes seem well-invested. You will see `alignas(64)`, as a test where I tried to force it to be misaligned ran slower. But this was by no means scientific.

So finally here are a number of benchmarks. These are based on yesterday's main branch. zmij is the mn branch, zmij_new is this patch.

With `-msse2`
image

With `-msse2 -mlzcnt`
image

With `-msse4.1`
image

With `-mavx2 -mlzcnt` (which implies SSE4.1, but uses a different insstruction encoding)
image

How the uptick at two digits is cured by `LZCNT` is a mystery, and the uptick at 17 digits is also something that I don't have an explanation for. It seems to affect all algorithms though. I'm not using a current version of dtoa-benchmark, so if there's a bug, I probably have it.

I made a quick'n'hacky implementation based on this of a 256bit version that converts the exponent along the way and that staves off another 0.5ns. So that may come soon.

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -594,7 +594,7 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
   __m128i mask128 = _mm_cmpgt_epi8(bcd, _mm_setzero_si128());
   uint16_t mask = _mm_movemask_epi8(mask128);
 #  if defined(__LZCNT__) && !defined(ZMIJ_NO_BUILTINS)
-  auto len = 32 - __lzcnt32(mask);
+  auto len = 32 - _lzcnt_u32(mask);
 #  else
   auto len = mask == 0 ? 0 : 64 - clz(mask);
 #  endif


zmij.cc
```diff
@@ -23,23 +23,30 @@ struct dec_fp {
...
-#ifndef ZMIJ_USE_NEON
-#  if ZMIJ_USE_SIMD && \
-      (defined(__ARM_NEON) || (defined(_MSC_VER) && defined(_M_ARM64)))
-#    define ZMIJ_USE_NEON 1
-#  else
-#    define ZMIJ_USE_NEON 0
+#if ZMIJ_USE_SIMD
+#  if defined(__ARM_NEON) || (defined(_MSC_VER) && defined(_M_ARM64))
+#    include <arm_neon.h>
+#    define ZMIJ_USE_SIMD_NEON
```
vitaut: Let's keep the `ZMIJ_USE_NEON` name and rename the other one to `ZMIJ_USE_SSE` to match.

zmij.cc
```diff
@@ -23,23 +23,30 @@ struct dec_fp {
...
+#  elif defined(__SSE2__)
+#    include <immintrin.h>
+#    define ZMIJ_USE_SIMD_SSE
```
vitaut: The convention is to define `ZMIJ_USE_<foo>` to 1 or 0 to simplify checks.

zmij.cc
```diff
@@ -467,7 +459,78 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
...
-#endif  // __ARM_NEON
+#  elif defined(ZMIJ_USE_SIMD_SSE)
+  uint32_t abbccddee = uint32_t(value / 100'000'000);
+  uint32_t ffgghhii = uint32_t(value % 100'000'000);
+  uint32_t a = abbccddee / 100'000'000;
+  uint32_t bbccddee = abbccddee % 100'000'000;
+
+  buffer = write_if_nonzero(buffer, a);
+
+  alignas(64) static const struct {
```
vitaut, line 558: Is alignas actually useful here? Isn't the default alignment for this struct already that?

zmij.cc
```diff
@@ -467,7 +459,78 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
...
+    __m128i one_hundred = _mm_set1_epi32(100);
+#  endif
+    __m128i divmod100 = _mm_set1_epi32((1 << 16) - 100);
+    __m128i div10 = _mm_set1_epi16((1 << 16) / 10 + 1);
+#  if defined(ZMIJ_USE_SIMD_SSE4_1)
+    __m128i divmod10 = _mm_set1_epi16((1 << 8) - 10);
+    __m128i bswap = _mm_set_epi8(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
+#  else // ZMIJ_USE_SIMD_SSE4_1
+    __m128i moddiv10 = _mm_set1_epi16(10 * (1 << 8) - 1);
+#  endif // ZMIJ_USE_SIMD_SSE4_1
```
vitaut: Is it possible to consolidate ifdefs or will it regress perf?

zmij.cc
```diff
@@ -467,7 +459,78 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
...
+
+  auto digits = _mm_or_si128(bcd, c.ascii0);
+
+  // determine number of leading zeros
+  __m128i mask128 = _mm_cmpgt_epi8(bcd, _mm_setzero_si128());
+  uint16_t mask = _mm_movemask_epi8(mask128);
+#  if defined(__LZCNT__) && !defined(ZMIJ_NO_BUILTINS)
+  auto len = 32 - __lzcnt32(mask);
+#  else
+  auto len = mask == 0 ? 0 : 64 - clz(mask);
```
vitaut, line 599: I guess we could introduce a 32-bit version of `clz`.

zmij.cc
```diff
@@ -467,7 +459,78 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
...
+  uint64_t bcd = to_bcd8(abbccddee % 100'000'000);
+  write8(buffer, bcd | zeros);
+  if (ffgghhii == 0) {
+    buffer += count_trailing_nonzeros(bcd);
+    return buffer - int(buffer - start == 1);
+  }
+  bcd = to_bcd8(ffgghhii);
+  write8(buffer + 8, bcd | zeros);
+  return buffer + 8 + count_trailing_nonzeros(bcd);
+#endif  // ARM_NEON
```
vitaut: The comment here is no longer correct, let's just remove it.

zmij.cc
```diff
@@ -23,23 +23,30 @@ struct dec_fp {
...
+#  if defined(__ARM_NEON) || (defined(_MSC_VER) && defined(_M_ARM64))
+#    include <arm_neon.h>
+#    define ZMIJ_USE_SIMD_NEON
+#  elif defined(__SSE2__)
+#    include <immintrin.h>
+#    define ZMIJ_USE_SIMD_SSE
+#    ifdef __SSE4_1__
+#      define ZMIJ_USE_SIMD_SSE4_1
+#    endif
+#  elif defined(_MSC_VER) && (defined(_M_X64) || (defined(_M_IX86_FP) && _M_IX86FP == 2))
```
AlexGuteniev: Per #63 , `_M_X64` -> `_M_AMD64`

TobiSchluter: Sure, no preference.

zmij.cc
TobiSchluter: according to MSVC it isn't 
image
That doesn't mean it's useful, I tried to brutally misalign it, and it ended up seeming minimally slower.

zmij.cc
TobiSchluter: The compiler loads the variables in some arbitrary order anyway, so it doesn't matter.

Edit: not arbitrary! highly optimized and well thought through!

zmij.cc
TobiSchluter: done

zmij.cc
TobiSchluter: done, but this means i have to define them to `0` in the unused cases which may or may not be to your liking.

zmij.cc
TobiSchluter: Done.

zmij.cc
TobiSchluter: The semantics re the zero check are different from our clz though. That's why I refrained from doing that.

Side note: as for the 32bit thing: the mask is actually 16bit, but this yields the shortest code (no size prefixes in the instructions, the 16bit version gives a 16bit result, so it needs the same number of conversions), and it was minimally faster than the 64bit version in my benchmarks (again, not very scientific).

vitaut:
Looks like you need to rebase.

TobiSchluter:
I'm colliding against this while rebasing. Doesn't this *not* include the header if `ZMIJ_USE_NEON` is defined on the command line?
```
#ifdef ZMIJ_USE_NEON
// Use the provided definition.
#elif !ZMIJ_USE_SIMD
#  define ZMIJ_USE_NEON 0
#elif defined(__ARM_NEON) || (defined(_MSC_VER) && defined(_M_ARM64))
#  include <arm_neon.h>
#  define ZMIJ_USE_NEON 1
#else
```

TobiSchluter:
@vitaut re my previous comment, I updated this to read
```
#ifdef ZMIJ_USE_NEON
// Use the provided definition
#elif defined(__ARM_NEON) || (defined(_MSC_VER) && defined(_M_ARM64))
#  define ZMIJ_USE_NEON 1
#else
#  define ZMIJ_USE_NEON 0
#endif

#ifdef ZMIJ_USE_SSE
// Use the provided definition
#elif defined(__SSE2__)
#  define ZMIJ_USE_SSE 1
#elif defined(_MSC_VER) && (defined(_M_AMD64) || (defined(_M_IX86_FP) && _M_IX86FP == 2))
#  define ZMIJ_USE_SSE 1
#else
#  define ZMIJ_USE_SSE 0
#endif

#ifdef ZMIJ_USE_SSE4_1
// Use the provided definition
#elif defined(__SSE4_1__)
#  define ZMIJ_USE_SSE4_1 1
#elif defined(_MSC_VER) && defined(__AVX__) // There's no way to check for /arch:SSE4.2 specifically
#  define ZMIJ_USE_SSE4_1 1
#else
#  define ZMIJ_USE_SSE4_1 0
#endif

#if ZMIJ_USE_NEON
#  include <arm_neon.h>
#endif

#if ZMIJ_USE_SSE4_1 && !ZMIJ_USE_SSE
#  error "User asked for SSE4.1 but SSE is not available or explicitly not requested."
#endif

#if ZMIJ_USE_SSE
#  include <immintrin.h>
#endif
```
which I think captures the intended logic. There was also an #include missing in test/benchmark.cc which I added to be able to build the whole thing.

Edit: I'll revise that quickly, actually the order between SSE2 and SSE4.1 should be reversed to really capture all legitimate options ...

Edit2: reworked the logic to avoid cases where SSE4.1 is requested but SSE2 explicitly disabled. Previously that would have disabled the SSE4.1 code. Sorry about the double-force-push. I noticed some stray whitespace just as I hit the enter key ...

Edit3: I realized that there are more ways the user can cause chaos and adapted this accordingly.

zmij.cc
vitaut: Nevermind, not sure why I thought the alignment would be 64 by default =).

vitaut:
> Doesn't this not include the header if ZMIJ_USE_NEON is defined on the command line?

Yeah, the version in `main` is not quite correct. Feel free to overwrite it with your version.

TobiSchluter:
> > Doesn't this not include the header if ZMIJ_USE_NEON is defined on the command line?
> 
> Yeah, the version in `main` is not quite correct. Feel free to overwrite it with your version.

Done. I guess it's consistent that I wasn't sure if I missed something in your version before taking like three tries to get it right myself.

zmij.cc
```diff
@@ -471,7 +486,76 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
...
+    __m128i ascii0 = _mm_set1_epi64x(zeros);
+  } c;
+
+// The BCD sequences are based on ones provided by Xiang JunBo.
+#  if ZMIJ_USE_SSE4_1
+  __m128i x = _mm_set_epi64x(bbccddee, ffgghhii);
+  __m128i y = _mm_add_epi64(x, _mm_mul_epu32(c.divmod10000, _mm_srli_epi64(_mm_mul_epu32(x, c.div10000), 40)));
+  __m128i z = _mm_add_epi64(y, _mm_mullo_epi32(c.divmod100, _mm_srli_epi32(_mm_mulhi_epu16(y, c.div100), 3))); // _mm_mullo_epi32 is SSE 4.1
+  __m128i big_endian_bcd = _mm_add_epi16(z, _mm_mullo_epi16(c.divmod10, _mm_mulhi_epu16(z, c.div10)));
+  __m128i bcd = _mm_shuffle_epi8(big_endian_bcd, c.bswap); // SSE3
```
AlexGuteniev: SSSE3 that is Supplemental Streaming SIMD Extension 3, which is different from the actual SSE3.
(The actual SSE3 seems to introduce no instructions useful for general purpose vectorization)

zmij.cc
```diff
@@ -471,7 +486,76 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
...
+
+  buffer = write_if_nonzero(buffer, a);
+
+  alignas(64) static const struct {
+    __m128i div10000 = _mm_set1_epi64x((1ull << 40) / 10000 + 1);
+    __m128i divmod10000 = _mm_set1_epi64x((1ull << 32) - 10000);
+    __m128i div100 = _mm_set1_epi32((1 << 19) / 100 + 1);
+    __m128i divmod100 = _mm_set1_epi32((1 << 16) - 100);
+    __m128i div10 = _mm_set1_epi16((1 << 16) / 10 + 1);
+#  if defined(ZMIJ_USE_SIMD_SSE4_1)
```
vitaut: The name also needs to be updated here: `ZMIJ_USE_SIMD_SSE4_1` -> `ZMIJ_USE_SSE4_1`.

zmij.cc
```diff
@@ -471,7 +486,76 @@ auto write_significand17(char* buffer, uint64_t value) noexcept -> char* {
...
+  __m128i bcd = _mm_shuffle_epi32(bcd_shuffled, _MM_SHUFFLE(0, 1, 2, 3));
+#  endif // ZMIJ_USE_SIMD_SSE4_1
+
+  auto digits = _mm_or_si128(bcd, c.ascii0);
+
+  // determine number of leading zeros
+  __m128i mask128 = _mm_cmpgt_epi8(bcd, _mm_setzero_si128());
+  uint16_t mask = _mm_movemask_epi8(mask128);
+#  if defined(__LZCNT__) && !defined(ZMIJ_NO_BUILTINS)
+  auto len = 32 - __lzcnt32(mask);
```
vitaut: Don't you need to include `x86intrin.h`?

zmij.cc
vitaut: Do you mean 64-bit `clz` here gives the shortest code and there is no benefit to using `__builtin_clz` instead of `__builtin_clzll`?

vitaut (changes requested):
A few more comments inline.

zmij.cc
TobiSchluter: Ugh yes, sorry about that

zmij.cc
TobiSchluter: I went by this: 
image
I think immintrin.h is the one big header that supercedes them all.

TobiSchluter: oh, i didn't notice the different name ... let me  check where that comes from ...

zmij.cc
TobiSchluter: The 16bit version doesn't give shorter code, even though the mask is naturally 16bits, was what I was trying to say. When I tried both the 32bit and 64bit versions the 32bit version benchmarked minimally faster, but I wouldn't call that a solid result.

As for the scode size, neither are two bytes very important, nor is an isolated demonstration very meaningful as it depends on the surrounding code, but here's a godbolt with the relevant logic from zmij isolated https://godbolt.org/z/z99odnxK4 

I'll have to figure out what is going on with the various lzcnts and will get back to you about that tonight.

vitaut:
I haven't tried more advanced configurations but can confirm an improvement with the defaults using gcc:

Before:
```
dragonbox: 30.61ns (27.96ns - 35.11ns)
zmij     : 24.28ns (22.85ns - 28.45ns)
```

After:
```
dragonbox: 30.61ns (27.89ns - 35.20ns)
zmij     : 22.81ns (21.05ns - 27.59ns)
```

vitaut:
Merged, thank you!

TobiSchluter:
@vitaut I really don't want to have even more social media accounts, but I saw your discussion about the static initializer.

 I also noted that clang manages to extract the initialization into static memory whereas gcc initializes at runtime.

I based this on Dougall's work, without being overly scientific. What I did test is removing the `static` keyword in front of the struct, i.e. using runtime intialization This leads to an 0.3ns slowdown.

I also tried if the alignment has an effect. Trying to deliberately misalign the struct (which has 16byte alignment, and which is 128bytes large i.e. exactly two cachelines), lead to a barely measurable slowdown (0.1ns at best), but under cache pressure that may be detrimental. In the assembly I observe that gcc loads one of the entries of the second half early. I didn't make much of it originally, but now I suspect that this is deliberate in order to populate the cache. It would probably be wise to see if that also happens without a struct. 

PS I want to point out who gave the assist for the biggest speedup, and we should also give props to clang for not detecting that clz pattern

Edit: for completeness, I also tried if an assembly barrier had any impact and it didn't.

AlexGuteniev:
I think that if it turns out that struct is desirable, but we want to avoid the runtime initialization, we can have struct of arrays of built-in integers, and use `_mm_load_si128` with reinterpret cast.

TobiSchluter:
```
  constexpr const uint16_t div10_u16 = (1 << 16) / 10 + 1;
  constexpr const uint16_t moddiv10_u16 = 10 * (1 << 8) - 1;

  alignas(64) static const struct {
    //__m128i div10k = _mm_set1_epi64x(div10k_sig);
    uint64_t div10k[2] = { div10k_sig, div10k_sig };
    //__m128i neg10k = _mm_set1_epi64x(::neg10k);
    uint64_t neg10k[2] = {::neg10k,::neg10k};
    //__m128i div100 = _mm_set1_epi32(div100_sig);
    uint32_t div100[4] = {div100_sig,div100_sig,div100_sig,div100_sig};
    //__m128i neg100 = _mm_set1_epi32(::neg100);
    uint32_t neg100[4] = {::neg100,::neg100,::neg100,::neg100};
    //__m128i div10 = _mm_set1_epi16((1 << 16) / 10 + 1);
    uint16_t div10[8] = {div10_u16, div10_u16, div10_u16, div10_u16, div10_u16, div10_u16, div10_u16, div10_u16, };
#  if ZMIJ_USE_SSE4_1
    __m128i neg10 = _mm_set1_epi16((1 << 8) - 10);
    __m128i bswap =
        _mm_set_epi8(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
#  else
    //__m128i hundred = _mm_set1_epi32(100);
    uint32_t hundred[4] = { 100, 100, 100, 100, };
    //__m128i moddiv10 = _mm_set1_epi16(10 * (1 << 8) - 1);
    uint16_t moddiv10[8] = { moddiv10_u16, moddiv10_u16, moddiv10_u16, moddiv10_u16, moddiv10_u16, moddiv10_u16, moddiv10_u16, moddiv10_u16, };
#  endif  // ZMIJ_USE_SSE4_1
    uint64_t zeros[2] = { ::zeros, ::zeros };
  } c;

  auto as128 = [](auto arg) { return *(__m128i*)arg; };
```

As a quick test, I changed it like this, wrapping all references in `as128` calls. https://godbolt.org/z/5rPeP94d8 It is performance neutral, at least on my laptop. gcc chooses to only keep three of the members in memory, synthesizing the others on the fly, but I can make no rhyme out of which ones it chooses (no 32bit one, but not all of the 16bit ones). I tried a memory barrier (asm that reads from the struct, but it had no impact, so maybe I messed it up.

TobiSchluter:
Since I'm mentioning the experiments that I did in this thread, let me add another unsuccessful one: if you remember all of this, my original plan was to use AVX 256bit lanes to also BCD convert the exponent at the same time as the significand, getting the exponent for free. This led to another 10% or so performance increase in my original attempt. I have revived that here https://github.com/TobiSchluter/zmij/tree/avx2. This is experimental, so not at all polished. It also doesn't contain a number of experiments I did including varying where I widen to 256bits, building the constants in different ways etc -- none of which had any significant impact on performance.

Unfortunately, it is slower. Or I should perhaps say fortunately, because that means that the extant implementation has become so much faster. To give you some idea of timings: on my AMD Ryzen the main branch at the time of this development clocked in at approx 9.5ns. With this patch we're somewhere near 10.5ns. With clang this becomes 15ns!

You may recall that the original reason I didn't pursue this was that it was so much slower on my laptop. This is still the case, after throttling my laptop to 80% to avoid any weirdness from heat-induced clock changes, the main branch clocks in at 35ns. This patch increases that to 49ns! Turns out that on the Tiger Lake core of my laptop `VZEROUPPER` is extremely expensive, and the compiler inserts it automatically at the end of any function that uses the upper parts of 256bit registers. So, 16 (or 32?) wide registers get zeroed, where three were used. This can be suppressed with the risky -mno-zeroupper. Lo and behold, after adding the flag, the laptop comes in at 39ns. Unfortunately, that's still slower than the main branch.

I had a good look through the AVX instruction sets and I didn't find anything that could be of much benefit. There is a 53bit multiply, but that operates on floating point ports and would thus incur latencies when moving back to the integer world. The only thing that seems possible is that the SSE2 path could be made faster by using some of the more advanced blending intrinsics to merge the div_100 and mod_100 lanes more compactly. I think the instruction set is more tuned to manually unrolled loops than any of the stuff we're doing here.

PR #60, opened 2026-01-06
Change csv.Writer to csv.writer

TobiSchluter:
Not sure if this is really the fix, but the script fails to run without this change. I don't know if it finishes, though :)

### DIFF
--- a/test/benchmark.py
+++ b/test/benchmark.py
@@ -32,7 +32,7 @@ def find_benchmark_exe(build: str):
     raise RuntimeError("benchmark executable not found")
 
 
-def benchmark_commit(sha: str, workdir: Path, writer: csv.Writer):
+def benchmark_commit(sha: str, workdir: Path, writer: csv.writer):
     run(["git", "checkout", "-q", sha], cwd=workdir)
 
     if not os.path.exists(workdir / "zmij.h"):


vitaut:
Thanks!

Note that `benchmark.py` is probably not what you want because it runs benchmark on **all** commits. I'll rename it to make it clearer. Normally you should run the benchmark program directly.

TobiSchluter:
Thanks, yeah, I was just curious how long this would take, and gave up after 3 commits :)

TobiSchluter:
@vitaut I forgot that I had actually started the script last night on my office PC. Here's what the plot looks like on amd64
image

I think it's obvious that this is not the main development platform :D

vitaut:
@TobiSchluter could you by any chance share the CSV file? There are a few interesting regressions that might be worth looking into.

TobiSchluter:
Should have done that right away! Will do tomorrow when I get back to the office

TobiSchluter:
Here you go @vitaut 
[results.csv](https://github.com/user-attachments/files/24512222/results.csv)

TobiSchluter:
I had ChatGPT generate this little script for me. According to it, the following are the worst regressions. I checked the first few manually to confirm that it worked correctly. The third one ... doesn't make me happy. It looks like it jumps down a few commits before and then back up again at this, so maybe the PC was really inspired during that little dip in execution time during that time? None of the changes look super-suspicious but it's all nano-second-optimizations ....

[find_culprit.py](https://github.com/user-attachments/files/24512258/find_culprit.py)

Commit: 041737f24d4676c6bfba26be00e28d691c148b5f
Increase: +4.77 ns (value: 18.05)
Log: 041737f Use a single shorter candidate

Commit: 341911f35ead9545a0ef95164f5015eedebbf2b1
Increase: +2.94 ns (value: 15.44)
Log: 341911f Add an experimental optimization

Commit: 45e7518553f864d35cce94c4a344698983b72277
Increase: +1.97 ns (value: 13.99)
Log: 45e7518 Merge pull request #16 from TobiSchluter/branchfree_exponent_pdp11

Commit: 4b77b3b49491e1f8f95e090199fd5d9f92c5307f
Increase: +1.86 ns (value: 15.20)
Log: 4b77b3b Improve portability.

Commit: a7e188d58ce5aecdb437b37db6aa00e03e299582
Increase: +1.75 ns (value: 13.36)
Log: a7e188d Remove leading zeros in subnormals

Commit: 92fe0848f0ac5f9b43941402d8e000deab4b133e
Increase: +1.12 ns (value: 13.29)
Log: 92fe084 Fix a regression in subnormals

Commit: d51d10c0a7ac2b72906efd02aabc78c47f3bac4d
Increase: +0.82 ns (value: 13.18)
Log: d51d10c Generate powers of 10 using method by Dougall Johnson

Commit: 24525c2bada8dbcc4780aed9d62bb48fd67c2cfc
Increase: +0.59 ns (value: 15.25)
Log: 24525c2 Remove wrong dragonbox result

Commit: 361d3cc2a1c602506b86390705fe83c1af13b5f8
Increase: +0.52 ns (value: 15.20)
Log: 361d3cc Fix typo in Schubfach reference

Commit: 71a732b68259eb4708c1c675b62cc976f224cc68
Increase: +0.52 ns (value: 15.20)
Log: 71a732b Add C++ usage example to README

TobiSchluter:
Here are the biggest decreases in execution time BTW

Commit: 881d25186222e36b3f9e75258ec4738b15adf5d2
Decrease: -4.92 ns (value: 13.13)
Log: 881d251 Fix a bound check

Commit: d797e1258e896fbaff0050dee834ee98978b6480
Decrease: -3.03 ns (value: 12.17)
Log: d797e12 Use BCD cascade for decimal to binary conversion.

Commit: b1f04dfbe3eea02fd40d06967e1a52eaaf43d607
Decrease: -1.74 ns (value: 13.71)
Log: b1f04df Minor tweaks

Commit: efc82ab765b00adccc5f096d682b705f082113e5
Decrease: -1.67 ns (value: 11.54)
Log: efc82ab Make exponent handling branch-free

Commit: 4f1f8bd0550896c7835cbcf934bb2d0b6caf96f6
Decrease: -1.52 ns (value: 13.18)
Log: 4f1f8bd Make exponent output branchless

Commit: 7b25f518dd70fb788da120ae09ef09b8fa1caa73
Decrease: -1.02 ns (value: 12.27)
Log: 7b25f51 Merge pull request #7 from TobiSchluter/bcd_for_significand

Commit: 37a7999409b3deca0b9c4d03e86bad6a679f74bf
Decrease: -0.95 ns (value: 13.00)
Log: 37a7999 Remove subnormals from the common path

Commit: f0f58fb576e983cedb93391a7d573899c33ba3eb
Decrease: -0.59 ns (value: 12.57)
Log: f0f58fb Remove remaining Javaisms

Commit: f5a2702e302e7eec6167c7647ae1d06ab072f18c
Decrease: -0.57 ns (value: 12.27)
Log: f5a2702 Use faster indexed loads on ARM - idea by Dougall Johnson

Commit: eacc7ccde730a5714b3c5d40b9b38507d263bddf
Decrease: -0.56 ns (value: 14.69)
Log: eacc7cc Update benchmark results and add dragonbox

TobiSchluter:
I ran it again over lunch, and the biggest changes appear for the same revisions as before, with the plot looking similar enough for me to trust it

image

[results_20260109.csv](https://github.com/user-attachments/files/24522561/results_20260109.csv)

TobiSchluter:
So https://github.com/vitaut/zmij/commit/45e7518553f864d35cce94c4a344698983b72277 and https://github.com/vitaut/zmij/commit/4b77b3b49491e1f8f95e090199fd5d9f92c5307f are the same patch, but it has been superseded by the new exponent handling.

Turns out that even with the new version of the code, gcc creates a jump instead of a conditional move. I assume my Intel laptop where i developed that patch handled the short, forward jump better than the AMD I'm running these benchmarks on. clang generates the proper conditional move

Here's an isolated godbolt test: https://godbolt.org/z/afhE7s7xE

Changing `(dec_exp >= 0) - 1` to `-(dec_exp < 0)` makes it 1ns faster, and using the ternary shaves off another 0.2ns! The original way of building a mask from the sign bit (`dec_exp >> 31`) also generates the optimal code. It's quite silly either way because the compiler obviously recognizes the "absolute value" idiom behind it. (clang generates identical code no matter which way the absolute value is expressed.)
image

I will see if I can reproduce that on my laptop and submit a PR.

ps and h*ly cow what is this in gcc's assembly 🙅 
image

Edit: gosh, I should stop looking at this trainwreck
image
WHY NO USE `lea rdx, [rdi + 2]`???

vitaut:
Top improvement and regression should have been one change resulting in a small overall improvement.

vitaut:
It is interesting that yy's optimization (https://github.com/vitaut/zmij/commit/341911f35ead9545a0ef95164f5015eedebbf2b1) is flagged as a regression. I guess it's mostly because the initial implementation was suboptimal. Most of it was recovered in https://github.com/vitaut/zmij/commit/b1f04dfbe3eea02fd40d06967e1a52eaaf43d607. Disabling it definitely makes perf much worse now.

vitaut:
Regressions related to subnormals don't matter anymore since the handling of subnormals no longer happens on the fast path.

vitaut:
The last three regressions are false positives since they don't have any code changes. This leaves the only potential regression: https://github.com/vitaut/zmij/commit/d51d10c0a7ac2b72906efd02aabc78c47f3bac4d.

PR #62, opened 2026-01-06
Lower the standard to C++14

### LINKED ISSUE #61: Reconsider standard version
https://github.com/vitaut/zmij/blob/73f00c356d02a77101ceec6bea288527e9bf95f2/CMakeLists.txt#L4

After 60dd279cdaf752198947391fef298d908154d0b8, the structured binding is gone, which, besides some attributes appear to be the only C++17 feature

I think MSVC will spit three warnings about `[[maybe_unused]]` but will compile in the default C++14 mode.

AlexGuteniev:
Low effort PR. Resolves #61.
 * Lower the standard of the library and tests to C++14
 * Keep benchmark standard C++17 for `from_chars`
 * Change `vector::data()` to what we did when we were young
 * Wrap `[[maybe_unused]]` into a macro. I honestly dunno what does it do here, maybe it is really unused?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.16)
 project(zmij CXX)
 
-set(ZMIJ_STANDARD cxx_std_17)
+set(ZMIJ_STANDARD cxx_std_14)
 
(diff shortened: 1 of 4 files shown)


vitaut:
Thank you!

PR #65, opened 2026-01-08
Ask for more Unicode

AlexGuteniev:
Doesn't build without this.

Not sure if this is a proper way, or should be centralized somehow.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -30,22 +30,16 @@ set(ZMIJ_CHECK_STANDARD cxx_std_20)
 
 add_library(fmt fmt/format.cc)
 if (MSVC)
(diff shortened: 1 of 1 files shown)


test/CMakeLists.txt
```diff
@@ -29,10 +29,16 @@ target_link_libraries(pow10-test gtest)
 set(ZMIJ_CHECK_STANDARD cxx_std_20)
 
 add_library(fmt fmt/format.cc)
+if (MSVC)
+    target_compile_options(fmt PRIVATE /utf-8)
```
vitaut: Let's make it PUBLIC and remove the float-check change.

vitaut (changes requested):
Thanks for the fix

vitaut:
> #define private public

LOL

PR #72, opened 2026-01-13
Peel off rightmost digit to gain cheaper division, avoid zero check, …

TobiSchluter:
…and make writing simpler.  A 10% performance gain.

This patch is derived from a remark Xiang JunBo made about significand lengths in one of the comments on the SIMD patches where he discussed the length of the significand and the last digit. This made me think about how to deal with the 17th digit in the SIMD code which writes 16 digits at once, and thus has to special case either the first or last digit. It occurred to me that because we are limited to numbers below 1e18 we can actually express division by ten cheaply with no shift in the assembly using mulhi (using the upper 64 bits of a wide multiplication).

If one peels off the rightmost digits and then uses the SIMD code to write the top 16digits at once, one finds that the first digit of these 16digits can be zero. This happens if the decimal significand has 16 digits -- we divided it by 10 to extract the last digit, but now we have only fifteen digits. And here is where the magic happens: because we allocate one extra byte in front of the buffer to be able to insert the decimal point, we can actually put the zero in that position, and everything falls in place without adding any branches.

This is a 10% speedup in my measurements.  On my laptop it looks like this
image

On my office PC it takes the benchmark time from 10.5ns to 9.4ns (with SSE2). So the 10ns barrier is broken also on x64.

A small additional gain is that since we know the 16digits can't all be zero, we no longer need a zero-check for `clz`. I believe the same optimization can be applied on NEON, but since I cannot test it, I only touched that code to deal with the updated signature of `write_significand17` -- with this change, the beginning of the buffer argument is moved by one byte to the left to accomodate for the potential leading '0'.

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -577,13 +577,9 @@ auto write_significand17(char* buffer, uint64_t value,
   buffer += 16 - ((zeroes != 0 ? clz(zeroes) : 64) >> 2);
   return buffer - int(buffer - start == 1);
 #elif ZMIJ_USE_SSE
-#  if ZMIJ_USE_INT128 == 1
-  constexpr int div10_exp = 64;
-  constexpr uint128_t div10_sig = (uint128_t(1) << div10_exp) / 10 + 1;
-  uint64_t digits_16 = uint64_t((value * div10_sig) >> div10_exp);
-#  else  // !ZMIJ_USE_INT128
-  uint64_t digits_16 = value / 10;
-#  endif  // ZMIJ_USE_INT128
+  // Divide by ten with a 64bit shift.  Note that (1 << 63) / 5 == (1 << 64) / 10
+  // but doesn't need an intermediate int128.
+  uint64_t digits_16 = use_umul128_hi64 ? umul128_hi64((1ull << 63) / 5 + 1, value) : value / 10;
   uint32_t last_digit = value - digits_16 * 10;
 
   // We always write 17 digits into the buffer, but the first one can be zero.


TobiSchluter:
@vitaut  I just noticed what appears to be a bug in the modified non-SIMD code, and I will try to figure out why this didn't trigger in the tests. I don't think I will touch anything but lines 653ff, so feel free to look at the other stuff if you're online already.

TobiSchluter:
@vitaut turns out the `#ifdef` circus I was struggling with the other day wasn't correct yet, and `zmij-no-simd-test` didn't successfully disable SIMD. This and the non-SIMD implementation of `write_significand17` are corrected by the added commits.

vitaut:
Thanks for another amazing PR!

I'll take a look in detail later but first a quick comment:

> we can actually express division by ten cheaply with no shift in the assembly using mulhi (using the upper 64 bits of a wide multiplication)

Not in this PR but if 64-bit multiplication is fast on your system, it might be worth trying to enable `use_umul128_hi64` to see if it makes things faster:

https://github.com/vitaut/zmij/blob/4086d94ef3eba4db036f0db5fe3fc337343fe655/zmij.cc#L218-L222

Currently it is only enabled on Apple Silicon where it does give a small win.

xjb714:
@TobiSchluter @vitaut
It can be proved that when the input floating-point number v is greater than 5e-324, the high 16 digit of the decimal result must not be zero. **double v; {d,k}=to_decimal(v); if(v>5e-324) : d/10!=0 is must exist.**
5e-324 is the smallest positive floating-point number within the double-precision range.
If we write the high 16 digits first, this way we can avoid the overhead of zero checking.

oscardssmith:
> Not in this PR but if 64-bit multiplication is fast on your system, it might be worth trying to enable use_umul128_hi64 to see if it makes things faster

This should likely be turned on unconditionally. 64 bit hi multiplication has been 4 cycles since at least Sandylake/Zen1 on the x86 side.

xjb714:
@TobiSchluter 
If the input floating-point number is 5e-324, the following value will be 0:
```uint64_t digits_16 = value / 10; // value=5```
The judgment should have been made based on the following conditions earlier to determine whether it was 5e-324.
```
double v;
if(*(u64*)&v == 1)write "5e-324" to buffer and return;
```

**Comment by TobiSchluter** [comment]
> Thanks for another amazing PR!
> 
> I'll take a look in detail later but first a quick comment:
> 
> > we can actually express division by ten cheaply with no shift in the assembly using mulhi (using the upper 64 bits of a wide multiplication)
> 
> Not in this PR but if 64-bit multiplication is fast on your system, it might be worth trying to enable `use_umul128_hi64` to see if it makes things faster:
> 
> https://github.com/vitaut/zmij/blob/4086d94ef3eba4db036f0db5fe3fc337343fe655/zmij.cc#L218-L222
> 
> Currently it is only enabled on Apple Silicon where it does give a small win.

I saw your post on mastodon, and noticed that you changed `(uint128_t(1) << 64) / 10` to `(1 << 63) / 5` and now I wish I had seen that! This was the piece of the puzzle that I was missing to actually use `umul128_hi64` and get rid of an `#ifdef`.

Actually, I had totally missed that Dougall had made the exact same multiplication optimization in `to_decimal`, that would have saved me some thinking!

BTW I'm not sure where the time saving actually comes from. From the assembly it seems that gcc chooses to jump over the length check if the last digit is non-zero. If that jump is well predicted it saves a number of instructions. It looks something like this:
```
       # long simd block until here
       test    ebp, ebp   # last digit == 0 ?
        jne     .L26
        vpxor   xmm1, xmm1, xmm1               # null? length determination
        vpcmpgtb        xmm0, xmm0, xmm1
        vpmovmskb       eax, xmm0
        mov     rdx, rax
        bsr     rax, rax
        cmp     edx, 1
        setne   dl
        movzx   edx, dl
        add     eax, edx
        cdqe
.L26:                                                           # code paths merge here
        movzx   edx, BYTE PTR [rsi+1]
```
Edit: That is with gcc. clang decides to not jump and just execute all the code always followed by `cmov`. I played around with getting the `cmov` out of gcc, but once I had it, it benchmarked slower (still faster than the old code, but not as much). I might have to install clang to get a comparison benchmark.

**Comment by TobiSchluter** [comment]
> @TobiSchluter @vitaut It can be proved that when the input floating-point number v is greater than 5e-324, the high 16 digit of the decimal result must not be zero. **double v; {d,k}=to_decimal(v); if(v>5e-324) : d/10!=0 is must exist.** 5e-324 is the smallest positive floating-point number within the double-precision range. If we write the high 16 digits first, this way we can avoid the overhead of zero checking.

Note that @vitaut normalizes subnormals (check `normalize()`) so the significand during printing is guaranteed to have only 16 or 17 digits. so the only overhed is determining whether one is dealing with 16 or 17 digits. But maybe that can be predicted as well?

**Comment by xjb714** [comment]
@TobiSchluter OK, I noticed that the `normalize` function has been newly added.

**Comment by TobiSchluter** [comment]
> Not in this PR but if 64-bit multiplication is fast on your system, it might be worth trying to enable `use_umul128_hi64` to see if it makes things faster:

@vitaut I guess this is enough and is in keeping with the policy you stated on mastodon:
     `uint64_t digits_16 = use_umul128_hi64 ? umul128_hi64((1ull << 63) / 5 + 1, value) : value / 10;`

Is there a reason you introduced `use_umul128_hi64` as a separate thing from `ZMIJ_USE_INT128`? I would -- perhaps naïvely, knowing only very little about ARM -- assume that a platform that supports `int128` (i.e. has `ZMIJ_USE_INT128 == 1`) has fast wide multiplication. To add to what @oscardssmith said, in the case of Intel, `mul` has produced a wide result since the 8086 days. Since the 80386 it even has `shrd`, a combined shift of `dx` and `ax` which is designed exactly to implement the multiply + shift trick.

Anyway, leaving aside history: on my AMD Ryzen 6500, with the above change, performance is not distinguishable between `use_umul128_hi64 == true` and `false`.

**Comment by vitaut** [comment]
> Is there a reason you introduced `use_umul128_hi64` as a separate thing from `ZMIJ_USE_INT128`?

I wasn't sure it is an improvement on other systems compared to 32-bit multiply and shift. Having it controlled separately makes it easy to measure while if it was folded into `ZMIJ_USE_INT128` you wouldn't be able to measure. If there is no difference then I suggest keeping 32-bit arithmetic. In any case it should definitely **not** be unconditional for obvious reasons.

**Comment by TobiSchluter** [comment]
Benchmark results with clang 18.3 on my AMD Ryzen 6500:
main: 10.1ns
this patch with umul128_hi64: 9.5ns
this patch without umul128_hi64: 10.2ns

gcc 14.2 for comparison:
main: 10.4ns
this patch with umul128_hi64: 9.50ns
this patch without umul128_hi64: 9.55ns <- probably same within noise

My worries that the gains with gcc may have been a benchmarking effect (less code due to an always correctly predicted jump) seem dispelled. clang does the full length evaluation and uses a `cmov` in the end instead of gcc's jump, and the gain is comparable. Interestingly, clang benefits from using 128bit divide-by-multiplication.

**Comment by TobiSchluter** [comment]
I updated the patch, basing it on #75 (the `#ifdef` fix). The only change besides separating out that fix in order to get rid of a merge conflict, is that I added the ternary using `use_umul128_hi64` from the previous comments given that my benchmarks with clang actually showed an improvement if the divide-by-mult trick was used. I didn't switch the default though, that is a  job for the captain of this ship.

**Comment by TobiSchluter** [comment]
BTW `split_tables` seems to be a small win on my laptop (1.5%). `LEA` can handle up to 8byte sized elements, so it's no surprise that there is a small gain in an isolated benchmark. I guess in an environment with memory pressure performance will very much depend on how many table entries end up being needed, and then the difference is small enough that it might come out on either side.

**Comment by vitaut** [comment]
Confirmed the improvement on my amd64 machine.

Before:
```
dragonbox: 30.61ns (27.95ns - 35.10ns)
zmij     : 21.09ns (20.43ns - 24.83ns)
```

After:
```
dragonbox: 30.60ns (27.87ns - 35.10ns)
zmij     : 19.88ns (19.62ns - 22.16ns)
```

vitaut:
Merged, thank you!

vitaut:
> I didn't switch the default though

If there is evidence that it improves perf on amd64 then we could default `use_umul128_hi64` to `ZMIJ_USE_INT128`. Then it will cover two main architectures with a fallback for other ones and test coverage from int128. A PR is welcome.

vitaut:
> BTW split_tables seems to be a small win on my laptop (1.5%).

Thanks for testing. On my test system (EPYC Milan) it caused a massive regression.

TobiSchluter:
Wow. I learn something new every day. Today: access patterns to a hot buffer that takes up 10% of L1 cache matter.

vitaut:
Peeling off the rightmost digit was a great idea because it opened more opportunities to reuse data computed earlier. Here's the initial attempt: https://github.com/vitaut/zmij/commit/aa2b909d49429adbfac77b4cca3d504492907560 (cleaned up in a later commit).

TobiSchluter:
Nice. I figured that the remainder of the code would know or could cheaply determine if there were 17 digits, so this is exactly what I was hoping for. 👍I was too stupid to figure it out myself, though :D

vitaut:
Unfortunately this method doesn't play well with recently added fixed format so I had to disable SSE on some paths (#86).

PR #75, opened 2026-01-14
Make ZMIJ_USE_SIMD-0 disable SIMD

TobiSchluter:
As the title says. This didn't work, probably since my changes to the `#ifdef`-chain. This fixes it.

Since this leads to merge conflicts and is independent from #72 where I originally included it out of necessity, I think it's best to make it a separate PR.


vitaut:
Thanks for the fix!

PR #76, opened 2026-01-16
Check int for zero explicitly to avoid compiler warning

AlexGuteniev:
I'm somewhat surprised if other compilers do not complain about implicit bool conversion


vitaut:
thanks

PR #77, opened 2026-01-19
Fix SSE path compilation

AlexGuteniev:
I _think_ this is the right way to define constants without intrisic types, and use pointer casts with loads:
 * The `__m128i` et al are not supposed to be accessed  using their data members. The data members are just debugger visualization aid.
 * Casting of SSE/AVX loads/stores parameter is fine, the intrinsics deal with suppressing strict aliasing rule too

I don't insist that this vision is necessarily right.

But MSVC does.

Let's fix MSVC compilation and do whatever was done before for other compilers.


vitaut:
Thank you!

PR #82, opened 2026-01-25
Use addition instead of OR to add zeros and thus allow use of LEA

TobiSchluter:
I don't believe my benchmarks here, but if I were to trust them this is a speed increase by 0.5ns in the `ZMIJ_NO_SIMD` case before the fixed format was added which breaks gcc's optimization. After it seems to be lost in the noise. It saves one assembly instruction in the output (not three as I hoped -- I wrote this before the fixed output was added which adds more opportunities). Check https://godbolt.org/z/86TT7MoGx which defines `OP` either as `+` or as `|` to allow for a straightforward comparison.

The idea here is that when we convert the BCD number into a string we used the logically correct bitwise `OR` operation, but using the `ADD` operation instead enables the use of the `LEA` instruction and thus (see around line 34 in the GCC output (which no longer inlines write_significand17,  probably because its shared with the fixed format), similarly in line 666 of clang's output)
```
        mov     r11, rdx
        or      r11, rcx
```` 
becomes simply
```
        lea     r11, [rdx+rcx]
```
I can't read arm assembly well, so I'm not going to make any guesses there.

(I originally pushed this to my fork with the last known revision where SSE works, but since this is irrelevant to this, I updated it to the main branch while writing the PR, hence the forced push.)

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -506,7 +506,7 @@ auto write_significand9(char* buffer, uint32_t value, bool has9digits) noexcept
     -> char* {
   buffer = write_if(buffer, value / 100'000'000, has9digits);
   uint64_t bcd = to_bcd8(value % 100'000'000);
-  write8(buffer, bcd | zeros);
+  write8(buffer, bcd + zeros);
   return buffer + count_trailing_nonzeros(bcd);
 }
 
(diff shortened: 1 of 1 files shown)


AlexGuteniev:
MSVC loves LEA too. It is compact, and also does not alter flags, so sometimes gives more room for instruction-level parallelism.

I'm wondering if it does not make things worse on Arm64 though,

vitaut:
Merged, thank you!

Issue #79, opened 2026-01-25
The C version does not build without 128-bit integers


AlexGuteniev:
Reasons:
1. Redeclaration 
`result` is being declared here:
https://github.com/vitaut/zmij/blob/3235751e4143459e579399f78720602470491548/zmij.c#L216-L224
and here again in the same scope:
https://github.com/vitaut/zmij/blob/3235751e4143459e579399f78720602470491548/zmij.c#L236-L237

2. 128-bit integer structure is lacking shifts operator:
First occurrence, but there are more:
https://github.com/vitaut/zmij/blob/3235751e4143459e579399f78720602470491548/zmij.c#L242-L244

I'm not sure if it is worth maintaining C version for MSVC, or just give up.

MasterDuke17:
I am attempting to use the C version here (https://github.com/MoarVM/MoarVM/pull/1986), but now I can't build it on linux (apple silicon or x86-64).

vitaut:
I think we should fix it. A PR would be welcome.

vitaut:
@MasterDuke17 what error do you get on Apple Silicon? An issue with the SSE version has already been reported and fixed elsewhere.

AlexGuteniev:
> I think we should fix it. A PR would be welcome.

I'm seeing that most of tests fail this time, Do they fail for you too?

vitaut:
AFAICS all tests pass on macOS/ARM and, with https://github.com/vitaut/zmij/commit/408ebfb0c70579895e12864aeb00561ef15729e3, amd64/Linux.

AlexGuteniev:
> I think we should fix it. A PR would be welcome.

### 🤬🤬🤬

The tests pass for me in Debug but not in Release!

AlexGuteniev:
> and, with [408ebfb](https://github.com/vitaut/zmij/commit/408ebfb0c70579895e12864aeb00561ef15729e3), amd64/Linux.

I forgot that CMake/ninja would stop further build if something fails (the C version in my case), so I was trying stale exe before this fix.

AlexGuteniev:
@vitaut does the C version work on amd64 for you now?

vitaut:
Yes by some definition of work: zmij-c-test passes.

MasterDuke17:
> [@MasterDuke17](https://github.com/MasterDuke17) what error do you get on Apple Silicon? An issue with the SSE version has already been reported and fixed elsewhere.

[...]

PR #83, opened 2026-01-25
Fix MSVC compilation of C version

AlexGuteniev:
Towards #79 
The C version passes without SIMD after that for me.
Still tests fail with SIMD.


AlexGuteniev:
> but I inadvertently removed it when updating the code.

It is still there actually differently named.
(I almost brought the old one back, which would have caused two of them present).

vitaut:
Merged, thanks for the fix!

vitaut:
> Still tests fail with SIMD.

Even with https://github.com/vitaut/zmij/commit/408ebfb0c70579895e12864aeb00561ef15729e3?

AlexGuteniev:
> > Still tests fail with SIMD.
> 
> Even with 408ebfb0c70579895e12864aeb00561ef15729e3?

Yes, ##84 fixes this

PR #85, opened 2026-01-25
Fix arm again

AlexGuteniev:
The previous fix had a typo, which was only discovered after trying to compile arm64


vitaut:
I think this is already fixed in https://github.com/vitaut/zmij/commit/ebdd42d93532ad77c2d2aeff5e04d6480e0004b1.

PR #87, opened 2026-01-25
Copy the fix from df35b9c for the C version

MasterDuke17:
Now in linux on my apple silicon laptop both gcc and clang build and pass the C test.


vitaut:
Thank you

PR #92, opened 2026-01-27
Fix warnings about func declarations in C version

MasterDuke17:
Fixes
```
3rdparty/zmij/zmij.c:993:1: warning: ‘inline’ is not at beginning of declaration [-Wold-style-declaration]
  993 | static const ZMIJ_INLINE unsigned char compute_exp_shift(int bin_exp,
      | ^~~~~~
3rdparty/zmij/zmij.c:993:8: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
  993 | static const ZMIJ_INLINE unsigned char compute_exp_shift(int bin_exp,
      |        ^~~~~
```


MasterDuke17:
> Thanks for the PR. Removing `const` makes sense but please don't reorder `static` and `ZMIJ_INLINE`. Despite somewhat confusing GCC diagnostic the order is actually correct.

So done.

vitaut:
Merged, thank you!

PR #94, opened 2026-01-28
Fix build of C test

TobiSchluter:
The test wasn't updated to the new interface.


TobiSchluter:
Ah well, that doesn't fix it. The test still fails, digit strings are cut off too early.

vitaut:
The test failures should be fixed by #95 and I also updated the API usage. Thanks!

PR #96, opened 2026-01-30
avx2 for +12% aka "Fck yo cache"

TobiSchluter:
image

After a lot of experimenting, I finally understood that avx2 only gains speed if one rephrased the problem in the way its designers designed it for: processing lots of data with table lookups. Well, adding a 40k table does indeed yield a +12% speed boost. Something absolutely elusive until this realization Since this is great and amazing and superb I decided to hardcode `-mavx2`.  Noone will ever want to use Apple after this, and L1 cache should be devoted to double conversion anyway.

There are amazing things one can do with AVX2. Faster BCD conversion with less cache ain't one of them, sadly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,8 @@ project(zmij CXX)
 
 set(ZMIJ_STANDARD cxx_std_14)
 
+add_definitions(-mavx2)
+
 add_library(zmij zmij.cc zmij.h)
 target_include_directories(zmij PUBLIC .)
 target_compile_features(zmij PRIVATE ${ZMIJ_STANDARD})


vitaut:
I thought you were trying to get rid of lookup tables, not add more =)

TobiSchluter:
digits2 is gone. Mission accomplished. 💪

AlexGuteniev:
This is concerning.

1. Unconditional AVX2 may be rough for some retrocomputing enthusiasts. Even modern operating systems, such as Windows 11, require SSE2 or SSE4.2, but not AVX2.
2. '-mavx2` for gcc/Clang will allow these compilers auto-vectorize with avx2. So opting out of SIMD will be potentially not opting out.
3. I heard that AVX2 gather is worse on some CPUs than on the others. I was pointed to this commit https://code.videolan.org/videolan/dav1d/-/commit/7072e79faa8b8f387960b4a738139c98c270277f as an example. On uops.info I only see bad numbers for E-cores of Alder Lake [here](https://uops.info/table.html?search=gatherdd%20&cb_lat=on&cb_tp=on&cb_uops=on&cb_ports=on&cb_RKL=on&cb_ADLP=on&cb_ADLE=on&cb_ZEN4=on&cb_measurements=on&cb_base=on&cb_avx2=on), but some number may be gathered before microcode update.

AlexGuteniev:
There's AVX2 `remove`/`unique` vectorization algorithm that uses `pshufb`/`vpermd`, it composes removal mask, and uses up to 4K lookup tables (4K is for `pshufb` of 16 mask 2K is for `vpermd` of 8 mask). I mean I got it approved to MSVC STL. But it is still far from a whooping 40K!

TobiSchluter:
I'm closing this to avoid further confusion over whether this is a serious proposal.

I tried a number of things using AVX2, including
- using table lookups for the two-digit blocks
- using the packing and unpacking mechanisms to streamline the SSE2 or SSE4.2 code paths
- using blend with mask to merge the two sets of digits in the SSE2 path
- using a 256bit register to transform the exponent along the way

None of them increased performance. Almost every operation in the BCD conversion depends on the result of the immediately preceding operation, so it's one long dependency chain. The chain can be split by handling the digits separately in for the DIV100 and MOD100 cases in the SSE2 path -- but since the result of that need to be merged again, this doesn't save anything. The only thing that doesn't fall along that dependency path is the length evaluation in the case of 17 digits, and unlike clang, gcc nicely separates that, breaking the dependency chain so that the final evaluation only needs to happen before the benchmark reads the output from memory (which it doesn't, which makes it an unrealistic saving).

Issue #102, opened 2026-02-10
MSVC C zmij.c compilation


jcol2:
The c implementation doesn't seem to compile using the msvc c compiler. Is there any intention to support this or am I just doing compiling it wrong lol. I use c unity builds a lot so support would be great. In addition to the errors running with '/W4' also generates a ton of warnings. Should I make a pr that fixes the errors / majority of the warnings?
Output of running `cl zmij.c`:
```
Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35220 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

zmij.c
zmij.c(161): error C2059: syntax error: '['
zmij.c(165): error C2059: syntax error: '['
zmij.c(177): error C2059: syntax error: '['
zmij.c(226): error C2063: 'uint32_t': not a function
zmij.c(228): error C2063: 'uint32_t': not a function
zmij.c(235): error C2063: 'uint32_t': not a function
zmij.c(235): error C2063: 'uint32_t': not a function
zmij.c(236): error C2374: 'result': redefinition; multiple initialization
zmij.c(219): note: see declaration of 'result'
zmij.c(237): error C2063: 'uint32_t': not a function
zmij.c(243): error C2088: built-in operator '>>' cannot be applied to an operand of type 'uint128_t'
zmij.c(243): warning C4033: 'umul128_hi64' must return a value
zmij.c(248): error C2440: 'type cast': cannot convert from 'uint128_t' to 'uint64_t'
zmij.c(248): error C2088: built-in operator '>>' cannot be applied to an operand of type 'uint128_t'
zmij.c(249): error C2088: built-in operator '>>' cannot be applied to an operand of type 'uint128_t'
zmij.c(249): error C2440: 'type cast': cannot convert from 'uint128_t' to 'uint64_t'
zmij.c(260): error C2088: built-in operator '>>' cannot be applied to an operand of type 'uint128_t'
zmij.c(266): error C2099: initializer is not a constant
zmij.c(268): error C2099: initializer is not a constant
zmij.c(269): error C2099: initializer is not a constant
zmij.c(270): error C2099: initializer is not a constant
zmij.c(273): error C2099: initializer is not a constant
zmij.c(294): error C2099: initializer is not a constant
zmij.c(295): error C2099: initializer is not a constant
zmij.c(296): error C2099: initializer is not a constant
zmij.c(297): error C2099: initializer is not a constant
zmij.c(300): error C2099: initializer is not a constant
[...]

PR #103, opened 2026-02-12
Fix MSVC C compilation errors on zmij.c

jcol2:
https://github.com/vitaut/zmij/issues/102

This fixes the errors when compiling zmij.c using the msvc c compiler.

I tested this by compiling zmij.c with `cl` in a build using `zmij_detail_write_double` and `zmij_detail_write_float`, as well as running this repos tests.

### DIFF
--- a/zmij.c
+++ b/zmij.c
@@ -1240,7 +1240,7 @@ static char* write_significand17(char* buffer, uint64_t value, bool has17digits,
 
   buffer += 16 - ((zeroes != 0 ? clz(zeroes) : 64) >> 2);
   return buffer;
-#elif 1
+#elif ZMIJ_USE_SSE
   uint32_t last_digit = value - value_div10 * 10;
 
   // We always write 17 digits into the buffer, but the first one can be zero.


zmij.c
```diff
@@ -1290,9 +1289,9 @@ static char* write_significand17(char* buffer, uint64_t value, bool has17digits,
   // is the last digit which we factored off. But in that case the number would
   // be printed with a different exponent that shifts the last digit into the
   // first position.
-  auto len = size_t(64) - clz(mask);  // size_t for native arithmetic
+  auto len = ((size_t)64) - clz(mask);  // size_t for native arithmetic
```
mmozeiko: You forgot this `auto`. Compile file with `clang-cl -W3 zmij.c` to catch these.

zmij.c
```diff
@@ -1221,47 +1237,31 @@ static char* write_significand17(char* buffer, uint64_t value, bool has17digits,
...
+  const __m128i neg10k = _mm_set1_epi64x(NEG10K);
+  const __m128i div100 = _mm_set1_epi32(DIV100_SIG);
+  const __m128i div10 = _mm_set1_epi16((1 << 16) / 10 + 1);
 #  if ZMIJ_USE_SSE4_1
-  const __m128i neg100 = _mm_load_si128((__m128i*)&consts.neg100);
-  const __m128i neg10 = _mm_load_si128((__m128i*)&consts.neg10);
-  const __m128i bswap = _mm_load_si128((__m128i*)&consts.bswap);
+  const __m128i neg100 = _mm_set1_epi32(NEG100);
+  const __m128i neg10 = _mm_set1_epi16((1 << 8) - 10);
+  const __m128i bswap =
```
jcol2: @mmozeiko Could you tell if this the correct intrinsic to use here? I've never touched simd before so I hope I'm not butchering this.

zmij.c
jcol2: Thank you!

zmij.c
jcol2: Specifically this: `      _mm_setr_epi8(15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0);`

mmozeiko: Yes, that is correct.
`_mm_setr_epi8(a,b,c,...,x,y)` is same byte order as `{a,b,c,...,x,y}` array in memory.
`_mm_set_epi8(a,b,c,...,x,y)` is the opposite: `{y,x,...,c,b,a}`

jcol2:
> Thanks for the PR!
> 
> I'll take a look later but as a general comment please don't rename any symbols.

For sure, I wasn't going to rename anything but some of the local variables like `neg10k` in `write_significand17` were colliding with global variables with the same name, which was causing errors. My not very bright solution was to just rename the global variables / macros to be all uppercase. Would be happy to change the names around to whatever suits your fancy!

zmij.c
```diff
@@ -1221,52 +1238,36 @@ static char* write_significand17(char* buffer, uint64_t value, bool has17digits,
...
-  } consts;
+  uint32_t abcdefgh = value_div10 / (uint64_t)1e8;
+  uint32_t ijklmnop = value_div10 % (uint64_t)1e8;
 
-  const __m128i div10k = _mm_load_si128((__m128i*)&consts.div10k);
-  const __m128i neg10k = _mm_load_si128((__m128i*)&consts.neg10k);
-  const __m128i div100 = _mm_load_si128((__m128i*)&consts.div100);
-  const __m128i div10 = _mm_load_si128((__m128i*)&consts.div10);
+  const __m128i div10k = _mm_set1_epi64x(div10k_sig);
+  const __m128i neg10k_wide = _mm_set1_epi64x(neg10k);
```
jcol2: Okie dokie, I've changed all the globals back to what they were named before, although some locals are named different to avoid an error: `neg10k_wide` `neg100_wide` and `zeros_wide`. Let me know what you want those to be called lol.

zmij.c
```diff
@@ -286,17 +297,18 @@ uint32_t umulhi_inexact_to_odd32(uint64_t x_hi, uint64_t _, uint32_t y) {
...
-static const int double_exp_bias = (1 << (double_num_exp_bits - 1)) - 1;
-static const int double_exp_offset = double_exp_bias + double_num_sig_bits;
+enum {
+  double_num_bits = 64,
+  double_num_sig_bits = DBL_MANT_DIG - 1,
+  double_num_exp_bits = double_num_bits - double_num_sig_bits - 1,
+  double_exp_mask = (1 << double_num_exp_bits) - 1,
+  double_exp_bias = (1 << (double_num_exp_bits - 1)) - 1,
+  double_exp_offset = double_exp_bias + double_num_sig_bits,
+};
```
vitaut, line 311: Why change to enum?

jcol2: When compiling with `cl`, code like this gives an `initializer is not a constant` error:
```c
 static int A = 1;
 static int B = A;
 ```
One way around this for `static int` variables is to make them an enum, since enums have type `int`. Then that sort of assignment works for some reason. In cases for other types I would make it a `#define`, both work just as well.

Comment (jcol2) [reply zmij.c:311]: I could change these enums over to #defines if you want, it might be more consistent that way. Up to you!

Comment (jcol2) [reply zmij.c:311]: Actually I just realized this `initializer is not a constant` error also occurs for just normal globals as well:
```c
int A = 1;
int B = A;
```
Kinda janky :(

zmij.c
```diff
@@ -110,6 +110,18 @@ static_assert(!ZMIJ_USE_SSE4_1 || ZMIJ_USE_SSE);
...
+#if defined(ZMIJ_ALIGNAS)
+// Use the provided definition
+#elif defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
+#  define ZMIJ_ALIGNAS(x) _Alignas(x)
+#elif ZMIJ_MSC_VER
+#  define ZMIJ_ALIGNAS(x) __declspec(align(x))
+#elif __GNUC__
+#  define ZMIJ_ALIGNAS(x) __attribute__((aligned(x)))
+#else
+#  error Failed to define ZMIJ_ALIGNAS
```
vitaut: Instead of an error this should be defined to nothing because alignas is just an optimization.

zmij.c
vitaut: Oh wow, MSVC is quite a mess. Anyway, I think enum is fine.

zmij.c
```diff
@@ -286,17 +297,18 @@ uint32_t umulhi_inexact_to_odd32(uint64_t x_hi, uint64_t _, uint32_t y) {
...
+  double_exp_mask = (1 << double_num_exp_bits) - 1,
+  double_exp_bias = (1 << (double_num_exp_bits - 1)) - 1,
+  double_exp_offset = double_exp_bias + double_num_sig_bits,
+};
 
 typedef uint64_t double_sig_type;
-static const double_sig_type double_implicit_bit = (double_sig_type)1
-                                                   << double_num_sig_bits;
+#define double_implicit_bit \
+  ((double_sig_type)((double_sig_type)1 << double_num_sig_bits))
```
vitaut: Let's avoid macros and just spell out the constant here:

```
static const double_sig_type double_implicit_bit = 0x10000000000000;
```

zmij.c
```diff
@@ -314,16 +326,18 @@ static int64_t double_get_exp(double_sig_type bits) {
...
+  float_exp_mask = (1 << float_num_exp_bits) - 1,
+  float_exp_bias = (1 << (float_num_exp_bits - 1)) - 1,
+  float_exp_offset = float_exp_bias + float_num_sig_bits,
+};
 
 typedef uint32_t float_sig_type;
-static const float_sig_type float_implicit_bit = (float_sig_type)1
-                                                 << float_num_sig_bits;
+#define float_implicit_bit \
+  ((float_sig_type)((float_sig_type)1 << float_num_sig_bits))
```
vitaut: same here

zmij.c
```diff
@@ -1030,15 +1046,16 @@ static inline const char* digits2(size_t value) {
...
-static const int div10k_exp = 40;
-static const uint32_t div10k_sig = (uint32_t)((1ull << div10k_exp) / 10000 + 1);
-static const uint32_t neg10k = (uint32_t)((1ull << 32) - 10000);
-static const int div100_exp = 19;
-static const uint32_t div100_sig = (1 << div100_exp) / 100 + 1;
-static const uint32_t neg100 = (1 << 16) - 100;
-static const int div10_exp = 10;
-static const uint32_t div10_sig = (1 << div10_exp) / 10 + 1;
-static const uint32_t neg10 = (1 << 8) - 10;
+#define div_10k_exp 40
```
vitaut: macro -> enum and keep the old name to match the reference implementation

zmij.c
```diff
@@ -1221,52 +1238,36 @@ static char* write_significand17(char* buffer, uint64_t value, bool has17digits,
...
-    m128i neg100 = splat32(::neg100);
-    m128i neg10 = splat16((1 << 8) - 10);
-    m128i bswap = m128i{pack8(15, 14, 13, 12, 11, 10, 9, 8),
-                        pack8(7, 6, 5, 4, 3, 2, 1, 0)};
-#  else
-    m128i hundred = splat32(100);
-    m128i moddiv10 = splat16(10 * (1 << 8) - 1);
-#  endif
-    m128i zeros = splat64(::zeros);
-  } consts;
```
vitaut, line 1244: I suggest keeping the struct but using designated initializers, something like:

```
static alignas(64) const struct {
  m128i div10k;
  m128i neg10k;
  m128i div100;
  m128i div10;
#if ZMIJ_USE_SSE4_1
  m128i neg100;
  m128i neg10;
  m128i bswap;
#else
  m128i hundred;
  m128i moddiv10;
#endif
  m128i zeros;
} consts = {
  .div10k = splat64(div10k_sig),
  .neg10k = splat64(neg10k),
  .div100 = splat32(div100_sig),
  .div10  = splat16((1u << 16) / 10 + 1),
#if ZMIJ_USE_SSE4_1
  .neg100 = splat32(neg100),
  .neg10  = splat16((1u << 8) - 10),
  .bswap  = /* some C initializer */,
#else
  .hundred  = splat32(100),
  .moddiv10 = splat16(10 * (1u << 8) - 1),
#endif
  .zeros = splat64(zeros),
};
```

vitaut (changes requested):

zmij.c
jcol2: Yeah that's a great idea, that was my first idea as well. The issue I ran into was the same thing as above (you can't assign variables, and function returns to static variables). So `splat64`, `div10k_sig`, etc all become macros. Of course when you make `neg10k` a macro it starts breaking struct definitions and other variables using that symbol, so I renamed it to `global_neg10k`. Anyway I've switched to a designated initializer, lmk what you think. From what mmozeiko told me the performance should be the same either way but I haven't verified.

Here's a godbolt so you know I'm not making this up 😆 
MSVC: https://godbolt.org/z/PevEczr75
Clang: https://godbolt.org/z/a1acPGqqa

jcol2: @vitaut

jcol2: Also I bumped the c++ version to 20 in order for the tests to compile the designated initializer in zmij.c.

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.16)
 project(zmij CXX)
 
-set(ZMIJ_STANDARD cxx_std_14)
+set(ZMIJ_STANDARD cxx_std_20)
```
vitaut: The standard should remain 14 here, please revert.

zmij.c
```diff
@@ -286,17 +297,17 @@ uint32_t umulhi_inexact_to_odd32(uint64_t x_hi, uint64_t _, uint32_t y) {
...
+  double_num_exp_bits = double_num_bits - double_num_sig_bits - 1,
+  double_exp_mask = (1 << double_num_exp_bits) - 1,
+  double_exp_bias = (1 << (double_num_exp_bits - 1)) - 1,
+  double_exp_offset = double_exp_bias + double_num_sig_bits,
+};
 
 typedef uint64_t double_sig_type;
-static const double_sig_type double_implicit_bit = (double_sig_type)1
-                                                   << double_num_sig_bits;
+static const double_sig_type double_implicit_bit = 0x10000000000000;
```
vitaut: Actually now that `double_num_sig_bits` is an enum, we don't need this change: https://www.godbolt.org/z/br7Gc5hW5

zmij.c
```diff
@@ -314,16 +325,17 @@ static int64_t double_get_exp(double_sig_type bits) {
...
+  float_num_exp_bits = float_num_bits - float_num_sig_bits - 1,
+  float_exp_mask = (1 << float_num_exp_bits) - 1,
+  float_exp_bias = (1 << (float_num_exp_bits - 1)) - 1,
+  float_exp_offset = float_exp_bias + float_num_sig_bits,
+};
 
 typedef uint32_t float_sig_type;
-static const float_sig_type float_implicit_bit = (float_sig_type)1
-                                                 << float_num_sig_bits;
+static const float_sig_type float_implicit_bit = 0x800000;
```
vitaut: Same here.

zmij.c
```diff
@@ -1030,17 +1044,23 @@ static inline const char* digits2(size_t value) {
...
-static const uint32_t div100_sig = (1 << div100_exp) / 100 + 1;
-static const uint32_t neg100 = (1 << 16) - 100;
-static const int div10_exp = 10;
-static const uint32_t div10_sig = (1 << div10_exp) / 10 + 1;
-static const uint32_t neg10 = (1 << 8) - 10;
+enum {
+  div10k_exp = 40,
+};
+#define global_div10k_sig ((uint32_t)((1ull << div10k_exp) / 10000 + 1))
+#define global_neg10k ((uint32_t)((1ull << 32) - 10000))
```
vitaut: These changes can also be reverted now (including name changes).

zmij.c
```diff
@@ -1084,26 +1106,23 @@ typedef struct {
...
-}
+#define splat64(x)                            \
+  {                                           \
+    .data = {(long long)(x), (long long)(x) } \
+  }
+#define splat32(x) splat64((uint64_t)(x) << 32 | (x))
+#define splat16(x) splat32((uint32_t)(x) << 16 | (x))
+#define pack8(a, b, c, d, e, f, g, h)                                     \
+  ((uint64_t)((h << 56) | (g << 48) | (f << 40) | (e << 32) | (d << 24) | \
+              (c << 16) | (b << 8) | (a)))
```
vitaut: Let's revert this and other SIMD-related changes, disable SIMD in MSVC and focus on minimal changes to make the C implementation compile in this PR.

vitaut:
Merged, thanks!

PR #106, opened 2026-02-22
Apply settings to the correct test

TobiSchluter:
By divine law every test introduces at least one new bug. Here's proof that the universe is still in balance.

The no-simd test was compiled with no-builtins on top of no-simd, and the no-builtins test was built with default settings. I resisted the temptation to enumerate all combinations of flags to retain the added test coverage this typo achieved, instead I only fixed the typo.


vitaut:
Good catch, thanks!

Issue #105, opened 2026-02-18
Rust Zmij implementation of writing double seems faster than C++


Antares0982:
This is just a discussion, which I think you might be interested in.

You might already know that Zmij has a Rust version: https://github.com/dtolnay/zmij, which I suspect that is faster than current C++ implementation in writing doubles. My test case comes from the well-known json benchmark file [canada.json](https://raw.githubusercontent.com/serde-rs/json-benchmark/refs/heads/master/data/canada.json). My machine is x86-64 (i13700k) Linux.

I have created a repo for testing this (comparing C/C++/Rust impls): https://github.com/Antares0982/zmij-playground. It seems that Rust impl use another way to do the fixed notation writer, which comsumes about 9% clockticks less than C++ version (The data is from Intel VTune Profiler). C/C++ codes are compiled using clang (a bit faster than GCC) using command line like

```
clang++ -std=c++17 -O3 -DNDEBUG -fPIC -fno-stack-protector -fomit-frame-pointer -I<...>/zmij-playground/zmij-cpp -shared -o <...>/zmij-playground/build/libs/libzmij_cpp.so <...>/zmij-playground/zmij-cpp/zmij.cc
```

The following are my test results. In the picture (float writer is not tested), the Rust part should be `zmij_detail_write_double` + `__memcpy_avx_unaligned_erms`, and C++ part is `zmij::detail::write<double>`.

<img width="1281" height="99" alt="Image" src="https://github.com/user-attachments/assets/3978d17e-435f-4fdf-bc11-98d903c2183b" />

```
=== double benchmark (5000 rounds × 91932 values, 100 warmup) ===
  C              459660000 calls       5483.99 ms total     11.93 ± 0.34 ns/call  (sink=7896585000)
  C++            459660000 calls       5112.92 ms total     11.12 ± 0.19 ns/call  (sink=7896585000)
  Rust           459660000 calls       4912.29 ms total     10.69 ± 0.23 ns/call  (sink=7896789000)
```

I wonder maybe we can learn from the Rust impl to do the fixed notation writer? The rust code achieve this by moving data around the buffer written by scientific notation writer
```rust
    if Float::NUM_BITS == 32 && (-6..=12).contains(&dec_exp)
        || Float::NUM_BITS == 64 && (-5..=15).contains(&dec_exp)
    {
        if length as i32 - 1 <= dec_exp {
            // 1234e7 -> 12340000000.0
            return unsafe {
[...]

PR #110, opened 2026-02-28
Improvement on SSE4.1 path of writing fixed doubles

Antares0982:
This PR is trying to reduce the read-after-write I mentioned in #105, which gives a 6% improvement if sse4.1 is enabled.

The idea is very straightforward: by deriving all possible values of `dec_exp` and `extra_digit`, generate a shuffle table and read a shuffle mask from that table to replace the original `bswap` mask used to obtain the `bcd`. This directly shuffles each byte into its correct position, which not only avoids adding many instructions but also avoids writing and then reading back from memory. The shuffle table was generated by this script [shift_table.py](https://gist.github.com/Antares0982/ccbedff2f90afc099db517476c5c70e9); its logic is consistent with the original code.

A simple observation shows that the first byte before the shuffle is always moved to buffer + 16, which lies outside the 16-byte range of an XMM register. Writing the low 8 bytes of the pre-shuffle data directly at buffer + 16 solves this problem, so you can write 16 + 8 bytes non-overlappingly.

```c++
  _mm_storeu_si128(reinterpret_cast<__m128i*>(buffer), digits);
  // only the byte at buffer + 16 is actually needed
  write8(buffer + 16, unshuffled_digits[0]);
```

Including the `extra_digit`, this writes at most 25 bytes in total.

This approach relies on` _mm_shuffle_epi8`, so it cannot be used in an SSE2-only build. NEON could theoretically implement the same scheme; if this implementation is acceptable I will attempt a NEON version.

Because some SSE-related code needed to be reused, I refactored the original code slightly by extracting several functions and the `sse_constants` struct;  the logic of the non‑SSE4.1 variant remains unchanged. Additionally, if I’m not mistaken, GitHub’s Ubuntu Actions runners actually support AVX2, so I added an SSE4.1 test option to the test CMake script and enabled SSE4.1 testing in the workflow.


vitaut:
Thanks for the PR. What's the performance effect of this change?

Antares0982:
> Thanks for the PR. What's the performance effect of this change?

About 5% ~ 6% improvement if sse4.1 is enabled. Using VTune it shows my implementation saves 5.3% clockticks
image

test/CMakeLists.txt
```diff
@@ -13,6 +15,12 @@ endfunction()
 add_zmij_test(zmij-test)
 target_sources(zmij-test PRIVATE pow10-test.cc)
 
+if(TEST_SSE4_1)
+  add_zmij_test(zmij-sse4-test)
+  target_compile_options(zmij-sse4-test PRIVATE -msse4.1)
+  target_compile_definitions(zmij-sse4-test PRIVATE ZMIJ_USE_SSE4_1=1)
+endif()
```
vitaut: I suggest checking if `-msse4.1` is supported with `check_cxx_compiler_flag` and removing the `TEST_SSE4_1` option.

zmij.cc
```diff
@@ -923,22 +947,111 @@ ZMIJ_INLINE auto to_decimal_fast(UInt bin_sig, int64_t raw_exp,
...
+  uint32_t mask = _mm_movemask_epi8(mask128) | (1u << 16);
+#  if defined(__BMI1__) && !defined(ZMIJ_NO_BUILTINS)
+  auto len = 16 - _tzcnt_u32(mask);
+#  else
+  auto len = 16 - ctz32(mask);
+#  endif
+
+  _mm_storeu_si128(reinterpret_cast<__m128i*>(buffer), digits);
+  // only the byte at buffer + 16 is actually needed
+  write8(buffer + 16, unshuffled_digits[0]);
```
vitaut: `unshuffled_digits[0]` is likely non-portable

TobiSchluter:
Nice, I tried this before myself and found no improvement, I wonder what I did wrong :)

You can cut the table size almost in half by writing it like
```constexpr uint128 double_sse4_shuffle_table[32] = {
    ZMIJ_PACK16(15, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 12, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 12, 11, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 12, 11, 10, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 12, 11, 10, 9, 9, 8, 7, 6, 5, 4, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 12, 11, 10, 9, 8, 8, 7, 6, 5, 4, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 12, 11, 10, 9, 8, 7, 7, 6, 5, 4, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 6, 5, 4, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 5, 4, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 4, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 3, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 2, 1),
    ZMIJ_PACK16(15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 1),
    ZMIJ_PACK16(15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0),
```
I.e. adding the first line of the extra_digits case at the top and then indexing like
```  const __m128i shuffler = _mm_load_si128(
      (const __m128i*)&double_sse4_shuffle_table[dec_exp + !extra_digit]);
  ```

BTW I learned only yesterday that gcc11 and clang12 added architecture levels for amd64 which clean up this mess a bit.

**Comment by TobiSchluter** [comment]
On my AMD Ryzen the version from the pull request is slower than the one at the branch point
```
for J in benchmark_antares/*; do echo +++ $J; for I in  1 2 3; do taskset -c 0-3 $J ; done ; done
+++ benchmark_antares/benchmark_branch_point
Mean of per-digit medians:
dragonbox: 16.99ns (15.09ns - 20.26ns) noisy
zmij     :  9.27ns ( 9.23ns -  9.36ns) noisy
Mean of per-digit medians:
dragonbox: 16.66ns (15.03ns - 20.13ns) noisy
zmij     :  9.36ns ( 9.23ns -  9.59ns) noisy
Mean of per-digit medians:
dragonbox: 16.53ns (14.95ns - 19.46ns) noisy
zmij     :  9.28ns ( 9.18ns -  9.49ns) noisy
+++ benchmark_antares/benchmark_sse4_improve
Mean of per-digit medians:
dragonbox: 16.59ns (15.19ns - 19.44ns) noisy
zmij     :  9.33ns ( 9.26ns -  9.47ns) noisy
Mean of per-digit medians:
dragonbox: 16.59ns (14.99ns - 19.71ns) noisy
zmij     :  9.49ns ( 9.29ns -  9.73ns) noisy
Mean of per-digit medians:
dragonbox: 16.68ns (15.39ns - 19.65ns) noisy
zmij     :  9.41ns ( 9.31ns -  9.73ns) noisy
```

I see now why I never see improvements: I'm working on the wrong CPU :D 

(branch point is 5c9d8d84e254373eb375223ee9baecc0ebcd0231, sse4_improve is 30b604ade6db1a1091ce651cdef58e9d2e476aee)

**Comment by TobiSchluter** [comment]
I thought a little bit more and I realized that the `extra_digit << 4` in the table lookup is actually harmful, so I went ahead and tried the simplification that I pointed out above. Even on my picky Ryzen it is as fast, maybe faster, than the version at the branch point. TBH I'm surprised by the gain:
```
+++ benchmark_antares/benchmark_sse4_improve_smaller_table
Mean of per-digit medians:
dragonbox: 16.78ns (14.92ns - 19.94ns) noisy
zmij     :  9.29ns ( 9.23ns -  9.34ns) noisy
Mean of per-digit medians:
dragonbox: 16.68ns (14.95ns - 19.44ns) noisy
zmij     :  9.28ns ( 9.21ns -  9.35ns) noisy
Mean of per-digit medians:
dragonbox: 16.80ns (15.39ns - 19.90ns) noisy
zmij     :  9.28ns ( 9.22ns -  9.36ns) noisy
```
I pushed the change here:
https://github.com/TobiSchluter/zmij/pull/new/sse4_improvement_improved

With GCC 13.3 the relevant part in front of the shuffle changes like this:
before:
```
        mov     ecx, r8d
        shl     ecx, 4
        or      ecx, edi
        shl     rcx, 4
        lea     rdx, [rip + (anonymous namespace)::double_sse4_shuffle_table]
        vpshufb xmm2, xmm1, xmmword ptr [rcx + rdx]
```
after
```
        add     r9d, edi
        shl     r9, 4
        lea     rcx, [rip + (anonymous namespace)::double_sse4_shuffle_table]
        vpshufb xmm2, xmm1, xmmword ptr [r9 + rcx]
```
ps this is just to confirm that I benchmarked with -msse4.1, even though the above assembly was generated with -mavx2. I prefer this for reading because the more modern instruction encoding makes for less inter-register moves and thus the assembly is much easier to follow.

**Comment by TobiSchluter** [comment]
I looked even closer and I didn't realize before that your code is only called in the fixed case, i.e. a small subset of all benchmark runs. So, the small observable slowdown in the benchmark on my AMD Ryzen is actually a large slowdown in the fixed case. I wasn't sure how to isolate this, but to get a very rough idea I did the following:

Since the non-fixed case is just the fixed case where the dot is inserted after the first digit one can call the fixed case code always by modifying `write()` like so:
```
  // Write significand.
  if (dec_exp >= -4 && dec_exp < compute_dec_exp(traits::digits + 1))
    return write_fixed<traits::num_bits>(buffer, dec.sig, dec_exp, extra_digit);
  char* start = buffer;
#if ZMIJ_USE_SSE4_1
  if (traits::num_bits == 64)
    buffer = write_fixed_double_sse4(buffer, dec.sig, 0, extra_digit);
  else {
    buffer =
        write_significand<traits::num_bits>(buffer + 1, dec.sig, extra_digit);
    start[0] = start[1];
    start[1] = '.';
  }
#else
    buffer =
        write_significand<traits::num_bits>(buffer + 1, dec.sig, extra_digit);
    start[0] = start[1];
    start[1] = '.';
#endif
```
Making this change slows the benchmark from ~9.4ns to ~10.8ns. (I understand that the gains from not copying in the fixed case are underestimated by this procedure, OTOH it always uses the same shuffle mask on that path. This is actually one of the things I had attempted and dismissed locally before.)

I think having a benchmark that only converts numbers in the range for the fixed format is needed to really understand the performance characteristics of this change.

**Comment by Antares0982** [comment]
@TobiSchluter I would very much like to know how to determine which of the following is causing the slowdown on your device:

1. Read-after-write on your AMD device has little impact. There may be on‑chip optimizations that mitigate it.
2. Newly added operations in this PR have increased the overall execution time.

I did observe a performance improvement on my Intel CPU; I tested using the playground I previously wrote: https://github.com/Antares0982/zmij-playground. There is a test case here that only covers the fixed writer case: https://github.com/Antares0982/zmij-playground/blob/main/test_input.txt. You might consider using this test case to run a benchmark.

**Comment by TobiSchluter** [comment]
That's cool, though I had to install cargo to be able to use it.

Your version:
```
  C              459660000 calls       5564.73 ms total     12.11 ± 0.62 ns/call  (sink=7896585000)
  C++            459660000 calls       4070.40 ms total      8.86 ± 0.47 ns/call  (sink=7896585000)
  Rust           459660000 calls       4277.64 ms total      9.31 ± 0.59 ns/call  (sink=7896789000)
```

The branch point:
```
  C              459660000 calls       5558.82 ms total     12.09 ± 0.69 ns/call  (sink=7896585000)
  C++            459660000 calls       4040.47 ms total      8.79 ± 0.49 ns/call  (sink=7896585000)
  Rust           459660000 calls       4460.75 ms total      9.70 ± 0.55 ns/call  (sink=7896789000)
```

Not sure why rust moves that much, but repeated tries seem to confirm that yours is slower or at least not faster.

**Comment by Antares0982** [comment]
> That's cool, though I had to install cargo to be able to use it.
> 
> Your version:
> 
> ```
> === double benchmark (5000 rounds × 91932 values, 100 warmup) ===
>   C              459660000 calls       5564.73 ms total     12.11 ± 0.62 ns/call  (sink=7896585000)
>   C++            459660000 calls       4070.40 ms total      8.86 ± 0.47 ns/call  (sink=7896585000)
>   Rust           459660000 calls       4277.64 ms total      9.31 ± 0.59 ns/call  (sink=7896789000)
> ```
> 
> The branch point:
> 
> ```
> === double benchmark (5000 rounds × 91932 values, 100 warmup) ===
>   C              459660000 calls       5558.82 ms total     12.09 ± 0.69 ns/call  (sink=7896585000)
>   C++            459660000 calls       4040.47 ms total      8.79 ± 0.49 ns/call  (sink=7896585000)
>   Rust           459660000 calls       4460.75 ms total      9.70 ± 0.55 ns/call  (sink=7896789000)
> ```
> 
> Not sure why rust moves that much, but repeated tries seem to confirm that yours is slower or at least not faster.

Could you share the assembly code generated in the `build/libs`, both branch point and this PR? I want to check if the same assembly code is slower on my machine too, so we can confirm that this is a platform-specific problem or not.

**Comment by Antares0982** [comment]
Also, a 0.7% slow down on AMD, 5.3% improvement on Intel, seems acceptable to me. But I'm not sure if the result differs if using another test data set.

**Comment by TobiSchluter** [comment]
Sure, here you go. Compiler is clang, not my usual gcc, because it defaulted to it.

[zmij_cpp_branch_point.s.txt](https://github.com/user-attachments/files/25680320/zmij_cpp_branch_point.s.txt)
[zmij_cpp_antares.s.txt](https://github.com/user-attachments/files/25680322/zmij_cpp_antares.s.txt)

**Comment by TobiSchluter** [comment]
@Antares0982  Sorry, I need to revisit the measurements, because the assembly doesn't contain what I think it should contain (no mention of unshuffled in ...antares.s is suspicious). Let me start from scratch

**Comment by TobiSchluter** [comment]
I was a bit more careful this time round and here are the results and the assembly files. I ran the benchmarks a few times, so the numbers are fairly stable for each.

TobiS version

```
  C              459660000 calls       5486.34 ms total     11.94 ± 0.53 ns/call  (sink=7896585000)
  C++            459660000 calls       3745.78 ms total      8.15 ± 0.60 ns/call  (sink=7896585000)
  Rust           459660000 calls       4193.01 ms total      9.12 ± 0.70 ns/call  (sink=7896789000)
```

Version in this PR
```
  C              459660000 calls       5571.01 ms total     12.12 ± 0.71 ns/call  (sink=7896585000)
  C++            459660000 calls       3746.20 ms total      8.15 ± 0.51 ns/call  (sink=7896585000)
  Rust           459660000 calls       4145.97 ms total      9.02 ± 0.48 ns/call  (sink=7896789000)
```

Unmodified playground
```
  C              459660000 calls       5531.70 ms total     12.03 ± 0.50 ns/call  (sink=7896585000)
  C++            459660000 calls       3941.73 ms total      8.58 ± 0.48 ns/call  (sink=7896585000)
  Rust           459660000 calls       4158.52 ms total      9.05 ± 0.43 ns/call  (sink=7896789000)
```

zmij main version 00de552
```
  C              459660000 calls       5520.57 ms total     12.01 ± 0.60 ns/call  (sink=7896585000)
  C++            459660000 calls       4016.07 ms total      8.74 ± 0.51 ns/call  (sink=7896585000)
  Rust           459660000 calls       4127.59 ms total      8.98 ± 0.45 ns/call  (sink=7896789000)
```
zmij branch point ae48b83
```
  C              459660000 calls       5480.40 ms total     11.92 ± 0.57 ns/call  (sink=7896585000)
  C++            459660000 calls       3902.90 ms total      8.49 ± 0.38 ns/call  (sink=7896585000)
  Rust           459660000 calls       4240.71 ms total      9.23 ± 0.77 ns/call  (sink=7896789000)
```

Interestingly there's a big regression on the main trunk, your version is faster and there is no speed gain for mine. This doesn't seem to match my results with the benchmark in the zmij repository at all. Maybe you can make more sense of this than me?

[zmij_cpp_assemblies.zip](https://github.com/user-attachments/files/25681254/zmij_cpp_assemblies.zip)

**Comment by Antares0982** [comment]
Strange. Does gcc reproduce the performance regression?

**Comment by TobiSchluter** [comment]
Yes, I thought of this as well on the way to the train. Will try tomorrow.

**Comment by Antares0982** [comment]
> I thought a little bit more and I realized that the `extra_digit << 4` in the table lookup is actually harmful, so I went ahead and tried the simplification that I pointed out above. Even on my picky Ryzen it is as fast, maybe faster, than the version at the branch point. TBH I'm surprised by the gain:
> 
> ```
> +++ benchmark_antares/benchmark_sse4_improve_smaller_table
> Mean of per-digit medians:
> dragonbox: 16.78ns (14.92ns - 19.94ns) noisy
> zmij     :  9.29ns ( 9.23ns -  9.34ns) noisy
> Mean of per-digit medians:
> dragonbox: 16.68ns (14.95ns - 19.44ns) noisy
> zmij     :  9.28ns ( 9.21ns -  9.35ns) noisy
> Mean of per-digit medians:
> dragonbox: 16.80ns (15.39ns - 19.90ns) noisy
> zmij     :  9.28ns ( 9.22ns -  9.36ns) noisy
> ```
> 
> I pushed the change here: https://github.com/TobiSchluter/zmij/pull/new/sse4_improvement_improved
> 
> With GCC 13.3 the relevant part in front of the shuffle changes like this: before:
> 
> ```
>         mov     ecx, r8d
>         shl     ecx, 4
>         or      ecx, edi
>         shl     rcx, 4
>         lea     rdx, [rip + (anonymous namespace)::double_sse4_shuffle_table]
>         vpshufb xmm2, xmm1, xmmword ptr [rcx + rdx]
> ```
> 
> after
> 
> ```
>         add     r9d, edi
>         shl     r9, 4
>         lea     rcx, [rip + (anonymous namespace)::double_sse4_shuffle_table]
>         vpshufb xmm2, xmm1, xmmword ptr [r9 + rcx]
> ```
> 
> ps this is just to confirm that I benchmarked with -msse4.1, even though the above assembly was generated with -mavx2. I prefer this for reading because the more modern instruction encoding makes for less inter-register moves and thus the assembly is much easier to follow.

Additional 2% improvement on my machine. I'll just merge this commit.

**Comment by vitaut** [comment]
> one can call the fixed case code always by modifying write() like so

I am not sure this is a meaningful benchmark. If you want to compare the fixed case only, it's better to restrict inputs accordingly (which is basically what zmij-playground does).

**Comment by Antares0982** [comment]
> [zmij_cpp_assemblies.zip](https://github.com/user-attachments/files/25681254/zmij_cpp_assemblies.zip)

There are 4 .s files but only `zmij_cpp_TobiS.s` contains `write_fixed_double_sse4`. The content in `zmij_c_pull_request.s` may not be correct. I'll test `zmij_cpp_TobiS.s` and `zmij_cpp_branch_point.s` anyway.

**Comment by Antares0982** [comment]
> TobiS version
> 
> ```
> === double benchmark (5000 rounds × 91932 values, 100 warmup) ===
>   C              459660000 calls       5486.34 ms total     11.94 ± 0.53 ns/call  (sink=7896585000)
>   C++            459660000 calls       3745.78 ms total      8.15 ± 0.60 ns/call  (sink=7896585000)
>   Rust           459660000 calls       4193.01 ms total      9.12 ± 0.70 ns/call  (sink=7896789000)
> ```
> 
> zmij branch point [ae48b83](https://github.com/vitaut/zmij/commit/ae48b835ee3e5ed77f1592b77c0b58058543697b)
> 
> ```
> === double benchmark (5000 rounds × 91932 values, 100 warmup) ===
>   C              459660000 calls       5480.40 ms total     11.92 ± 0.57 ns/call  (sink=7896585000)
>   C++            459660000 calls       3902.90 ms total      8.49 ± 0.38 ns/call  (sink=7896585000)
>   Rust           459660000 calls       4240.71 ms total      9.23 ± 0.77 ns/call  (sink=7896789000)
> ```

I compiled `zmij_cpp_branch_point.s` to `libzmij-old`, `TobiS.s` to `libzmij-new`, and the latest code in this PR to `libzmij-cpp-new`. My clang version is 21.1.8. Seems similar to your result.
```
$ ./build/any_dtoa_benchmark ./build/libs/libzmij-old.so:zmij_detail_write_double ./build/libs/libzmij-new.so:zmij_detail_write_double ./build/libs/libzmij-cpp-new.so 
Reading values from: test_input.txt
Loaded 91932 values.

  Loaded: zmij-old
  Loaded: zmij-new
  Loaded: zmij-cpp-new

  zmij-old                   459660000 calls       4459.04 ms total      9.70 ± 0.40 ns/call  (sink=7896585000)
  zmij-new                   459660000 calls       4385.38 ms total      9.54 ± 0.07 ns/call  (sink=7896585000)
  zmij-cpp-new               459660000 calls       4344.15 ms total      9.45 ± 0.06 ns/call  (sink=7896585000)
```

**Comment by Antares0982** [comment]
BTW you may want to pull the latest zmij-playground. The new `any_dtoa_benchmark` can work with any dylib exporting a writer function. `tools/` contains some useful scripts for testing purpose.

**Comment by TobiSchluter** [comment]
@Antares0982 I think what happened was that when I accidentally / automatically hit F7 to build in VS Code, it wasn't able to build (because cargo wasn't yet in path), but it changed the build config so that if I wasn't careful to re-configure it would forget about the SSE4.1 setting.  Anyway, it seems to be faster also with gcc on more careful testing.

I added a few more changes on top of yours on my branch, they are all performance-neutral on my AMD, both in your benchmark and the one in the zmij. nevertheless I would be curious how it fares on your PC. As before it's here: https://github.com/TobiSchluter/zmij/pull/new/sse4_improvement_improved

The trivial things it changes:
1. clean up the shuffle table: instead of repeating the digits in an inscrutable pattern, use -1 for the field where the dot goes
2. it moves some stuff around to simplify the calculation where the dot is to be inserted (no `start *` needed), it is basically given by the index into the shuffle table
3. bswap is not used in the function, so it can be removed

And the potentially performance-impacting changes (not on my machine, where it appears neutral or maybe a very small enhancement)
1. since we updated the shuffle table, we can also use it as blend table to insert the dot while still doing SSE stuff
2. don't use `write8` to output the lone character at the end but just write that byte. I'm not sure if that is supposed to be faster or not, it seems to make no difference (NB our buffer is 25 bytes, so there is no write-beyond-end risk with `write8` which was my original fear Edit: see below!)

I'm happy to put this in an independent PR but feel free to merge if you think it is sound.

I thought about eliminating bswap and using the last entry in the shuffle table for that purpose, but since that is only available when not optimizing for size it seemed not worth the effort.

Edit: **actually** there is an overflow: sign + extra_digit + 16 + 8 = 26. I forgot about the sign.

**Comment by Antares0982** [comment]
> 2. don't use `write8` to output the lone character at the end but just write that byte. I'm not sure if that is supposed to be faster or not, it seems to make no difference (NB our buffer is 25 bytes, so there is no write-beyond-end risk with `write8` which was my original fear Edit: see below!)

Current `_mm_cvtsi128_si64` is a no-op, write8 generates single movq. But we can change this to a `write4` (using `_mm_cvtsi128_si32 `), which generates a movd. I believe that write only 1 or 2 bytes generates more u-op.

**Comment by TobiSchluter** [comment]
Agreed, write4 sounds safe

**Comment by TobiSchluter** [comment]
I changed my branch to use
```
  uint32_t trailing_digit = _mm_cvtsi128_si32(unshuffled_digits);
  memcpy(buffer + 16, &trailing_digit, 4); // only need the lowest byte
```
since there is no `write4`, and I don't think it's worth adding it when there's only one place where we don't even have to advance the buffer pointer.

Here's the comparison: 8bits on the left, 32bits on the right. 8bits has to move via EDX which is more work indeed (and moves the next access to EDX closer to its use)
image

**Comment by Antares0982** [comment]
> 1. since we updated the shuffle table, we can also use it as blend table to insert the dot while still doing SSE stuff

When we have the blendv:
```
  sse4-improve               459660000 calls       4359.31 ms total      9.48 ± 0.42 ns/call  (sink=7896585000)
  sse4_improvement_improved   459660000 calls       4496.90 ms total      9.78 ± 0.32 ns/call  (sink=7896585000)
```

Seems `_mm_blendv_epi8` introduces some extra latency. If related commits not picked, everything is fine.

I'll just pick some commits from your branch if you are OK with that.

**Comment by Antares0982** [comment]
Also I think 0x80 is a better value for generating zero byte in `_mm_shuffle_epi8`, so I replaced -1 with it. Any u8 will not change the result though.

**Comment by Antares0982** [comment]
If you guys are Okay I think it's time for another review

**Comment by TobiSchluter** [comment]
I feared that blendv performance would vary across different CPU cores. (The reason it was appealing enough for me to propose it is that I've been experimenting with avx2 to convert multiple numbers simultaneously, where one wants to keep as much of the work within SIMD as possible. Not sure if that's something for fmt, but it's fun.)

I like -1 because it stands out nicely, but I'm obviously fine with any way of setting the high bit. Same for cherry-picks of course.

**Comment by vitaut** [comment]
Merged, thanks!

Just as another data point, I confirmed the improvement on my AMD EPYC machine:

main:
```
=== float benchmark (5000 rounds × 91932 values, 100 warmup) ===
  C                          459660000 calls       7407.93 ms total     16.12 ± 0.05 ns/call  (sink=6036043800)
  C++                        459660000 calls       6596.72 ms total     14.35 ± 0.05 ns/call  (sink=4160472900)
  Rust                       459660000 calls       8440.70 ms total     18.36 ± 0.08 ns/call  (sink=4160850300)
  xjb                        459660000 calls       5265.61 ms total     11.46 ± 0.07 ns/call  (sink=4160850300)

  C                          459660000 calls      10254.25 ms total     22.31 ± 0.09 ns/call  (sink=7896585000)
  C++                        459660000 calls       7337.07 ms total     15.96 ± 0.16 ns/call  (sink=7896585000)
  Rust                       459660000 calls       7798.62 ms total     16.97 ± 0.05 ns/call  (sink=7896789000)
  xjb                        459660000 calls       9542.40 ms total     20.76 ± 0.10 ns/call  (sink=7896789000)
```

PR:
```
=== float benchmark (5000 rounds × 91932 values, 100 warmup) ===
  C                          459660000 calls       7407.54 ms total     16.12 ± 0.06 ns/call  (sink=6036043800)
  C++                        459660000 calls       6502.48 ms total     14.15 ± 0.29 ns/call  (sink=4160472900)
  Rust                       459660000 calls       8440.73 ms total     18.36 ± 0.10 ns/call  (sink=4160850300)
  xjb                        459660000 calls       5262.07 ms total     11.45 ± 0.05 ns/call  (sink=4160850300)

  C                          459660000 calls      10277.08 ms total     22.36 ± 0.18 ns/call  (sink=7896585000)
  C++                        459660000 calls       6174.09 ms total     13.43 ± 0.05 ns/call  (sink=7896585000)
  Rust                       459660000 calls       7794.48 ms total     16.96 ± 0.06 ns/call  (sink=7896789000)
  xjb                        459660000 calls       9550.76 ms total     20.78 ± 0.08 ns/call  (sink=7896789000)
```

There is a minor regression on zmij's own benchmark that mostly exercises exponential format but I think we can live with it.

Funny enough, `double` conversion is now faster that `float` conversion. We might wan't to apply SIMD optimizations to the latter as well.

PR #113, opened 2026-03-14
Avoid disallowed conversion from unsigned int to int

TobiSchluter:
`_lzcnt_u32` returns an `unsigned int` so `len` becomes `unsigned int` with `-march=x86-64-v3` where LZCNT is available. The constructor for the return value on the next line expects int. Fixed by making the type explicit in both `#ifdef` branches.

WRT my other PR, I'm still trying to figure out if your new table (also included when optimizing for size BTW) can be leveraged or integrated for the SSE code.


vitaut:
Merged, thanks!

vitaut:
> I'm still trying to figure out if your new table (also included when optimizing for size BTW) can be leveraged or integrated for the SSE code.

I was also thinking if the shuffle table can be merged into it.

PR #114, opened 2026-03-15
Improvement on NEON path of writing fixed doubles

Antares0982:
This is a continuation of #110 .

The only tricky part is that we have to reverse the hi-lo of `ffgghhii_bbccddee_64`, because the original order of unshuffled bytes cannot be used to calculate length. 

Also the fast path avoids reading `dec_exp_formats`.

Benchmark on my Apple M4, using `any_dtoa_benchmark`:
```
  zmij-neon-improve          459660000 calls       2279.64 ms total      4.96 ± 0.09 ns/call  (sink=7896585000)
  zmij                       459660000 calls       2490.61 ms total      5.42 ± 0.06 ns/call  (sink=7896585000)
```
About 8.4% improvement.

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -779,10 +779,10 @@ ZMIJ_INLINE auto get_double_unshuffled_digits_neon(char* buffer, uint64_t value,
   uint64x1_t ffgghhii_bbccddee_64;
   if (reverse_hi_lo) {
     // Used in write_fixed_double_fast.
-    // Only in this order we can use ctz for fast length calculation
-    // using unshuffled digits, which hides the latency of shuffle,
-    // And also let the 17th byte be always at index 0 before shuffle.
-    // The real order is: bbccddee_ffgghhii
+    // Only this ordering lets us use ctz for fast length calculation
+    // on unshuffled digits, hiding shuffle latency and ensuring
+    // the 17th byte is always at index 0 before shuffling.
+    // The actual order is: bbccddee_ffgghhii
     ffgghhii_bbccddee_64 = {(uint64_t(bbccddee) << 32) | ffgghhii};
   } else {
     ffgghhii_bbccddee_64 = {(uint64_t(ffgghhii) << 32) | bbccddee};


test/benchmark.cc
```diff
@@ -54,8 +53,7 @@ auto get_random_digit_data(int digit) -> const double* {
         // Limit the number of digits.
         char buffer[64] = {};
         fmt::format_to_n(buffer, sizeof(buffer), "{:.{}}", d, digit);
-        d = 0;
-        std::from_chars(buffer, buffer + strlen(buffer), d);
+        d = strtod(buffer, nullptr);
```
vitaut, line 56: Why switch to `strtod`?

zmij.cc
```diff
@@ -1137,6 +1222,9 @@ auto write(Float value, char* buffer) noexcept -> char* {
 #if ZMIJ_USE_SSE4_1 && !ZMIJ_OPTIMIZE_SIZE
     if (traits::num_bits == 64 && dec_exp >= 0)
       return write_fixed_double_sse4(buffer, dec.sig, dec_exp, extra_digit);
+#elif ZMIJ_USE_NEON && !ZMIJ_OPTIMIZE_SIZE
+    if (traits::num_bits == 64 && dec_exp >= 0)
+      return write_fixed_double_neon(buffer, dec.sig, dec_exp, extra_digit);
 #endif
```
vitaut: Let's merge `write_fixed_double_sse4` and `write_fixed_double_neon` into a single function with conditional compilation used internally and simplify this logic to something like

```
#if ZMIJ_USE_SIMD && !ZMIJ_OPTIMIZE_SIZE
    if (traits::num_bits == 64 && dec_exp >= 0)
      return write_fixed_double(buffer, dec.sig, dec_exp, extra_digit);
#endif
```

zmij.cc
```diff
@@ -725,34 +740,10 @@ ZMIJ_INLINE auto get_double_significand_bcd_unshuffled_sse(
...
-  // Digits/pairs of digits are denoted by letters: value = abbccddeeffgghhii.
-  uint32_t abbccddee = uint32_t(value / 100'000'000);
-  uint32_t ffgghhii = uint32_t(value % 100'000'000);
-  write_if(buffer, abbccddee / 100'000'000, extra_digit);
-  uint64_t hi = to_bcd8(abbccddee % 100'000'000);
-  if (ffgghhii == 0) return {{hi + zeros, zeros}, count_trailing_nonzeros(hi)};
-  uint64_t lo = to_bcd8(ffgghhii);
-  return {{hi + zeros, lo + zeros}, 8 + count_trailing_nonzeros(lo)};
-#elif ZMIJ_USE_NEON
+template <bool reverse_hi_lo = false>
```
vitaut, line 745: A brief comment explaining the effect of `reverse_hi_lo` would be good.

zmij.cc
```diff
@@ -925,6 +960,56 @@ auto write_fixed_double_sse4(char* buffer, uint64_t dec_sig, int dec_exp,
 }
 #endif
 
+#if ZMIJ_USE_NEON && !ZMIJ_OPTIMIZE_SIZE
+// clang-format off
```
vitaut: I don't think we need to disable clang-format here.

zmij.cc
```diff
@@ -925,6 +960,56 @@ auto write_fixed_double_sse4(char* buffer, uint64_t dec_sig, int dec_exp,
...
+constexpr uint8x16_t shuffle_table[] = {
```
vitaut: Other than type, is it different from the SSE table? Would be nice to have them in the same format.

zmij.cc
```diff
@@ -925,6 +960,56 @@ auto write_fixed_double_sse4(char* buffer, uint64_t dec_sig, int dec_exp,
...
+    uint8x16_t{0x80, 0x0f, 0x0e, 0x0d, 0x0c, 0x0b, 0x0a, 0x09, 0x08, 0x07, 0x06, 0x05, 0x04, 0x03, 0x02, 0x01},
```
vitaut: I think you can omit `uint8x16_t` here.

vitaut (changes requested):
Thanks for the PR!

test/benchmark.cc
Antares0982: from_chars doesn't compile on my machine and CI

zmij.cc
Antares0982: There is a comment at where it is used. I'll move them here

Antares0982:
The shuffle table is moved to a struct, and initialized by a constexpr constructor. It is strange that performance improved even on SSE4 path after the refactor :laughing: 

```
  zmij                       459660000 calls       4462.58 ms total      9.71 ± 0.12 ns/call  (sink=7896585000)
  zmij-neon-improve          459660000 calls       4300.47 ms total      9.36 ± 0.16 ns/call  (sink=7896585000)
```

Antares0982:
Can we do a re-review now?

vitaut:
Merged, thank you!

PR #115, opened 2026-03-18
Add windows ci, Fix node20 deprecation warning

Antares0982:
Add a ci for MSVC compiler, its behavior may differ from gcc/clang.

I encountered some compile errors on my Windows but it didn't show up on ci :/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,15 @@ jobs:
       matrix:
         os: [ubuntu-24.04, macos-15]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - run: cmake -B build
       - run: cmake --build build -j
       - run: ctest --test-dir build --output-on-failure
+
+  build-msvc:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v5
+      - run: cmake -B build
+      - run: cmake --build build --config Debug -j
+      - run: ctest --test-dir build -C Debug --output-on-failure


Antares0982:
On my PC there are a lot of errors: https://pastebin.com/V3mFHrRv
If you want a fix I can push some changes on this branch (the errors didn't occur in ci and I don't know why. I'm not very familiar with MSVC).

.github/workflows/ci.yml
```diff
@@ -7,7 +7,15 @@ jobs:
...
       - run: cmake --build build -j
       - run: ctest --test-dir build --output-on-failure
+
+  build-msvc:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v5
+      - run: cmake -B build
+      - run: cmake --build build --config Debug -j
+      - run: ctest --test-dir build -C Debug --output-on-failure
```
vitaut, line 21: Do we need to specify config here. Why not add windows to the os matrix above and reuse commands?

vitaut (changes requested):
Thanks for the P!

.github/workflows/ci.yml
```diff
@@ -7,7 +7,15 @@ jobs:
       matrix:
         os: [ubuntu-24.04, macos-15]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
```
vitaut, line 10: Is this related to the node deprecation warning?

vitaut:
Regarding Windows fixes, they are welcome but as a separate PR.

.github/workflows/ci.yml
Antares0982: For ctest we have to specify -C Debug here. See https://github.com/Antares0982/zmij/actions/runs/23397295670/job/68062662012

.github/workflows/ci.yml
Antares0982: Yes.
```
The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026.
```

vitaut:
Merged, thank you!

PR #119, opened 2026-04-08
Added correct fallback for 32 bit systems.

IRainman:


zmij.cc
```diff
@@ -14,12 +14,13 @@ struct dec_fp {
   bool negative;
 };
 }  // namespace zmij
+
+#  include <stddef.h>  // size_t
+#  include <string.h>  // memcpy
```
vitaut: This change looks incorrect, please revert.

vitaut (changes requested):

zmij.cc
IRainman: It's a include optimization because if the header exists, this include isn't needed.

IRainman: Now it's reverted

vitaut:
Merged, thanks

Issue #99, opened 2026-02-08
Optimize float output


xjb714:
The process of converting decimal numbers to BCD code can also be optimized. Use the SIMD instruction set. Please refer to https://github.com/xjb714/xjb/blob/main/src/ftoa_json.cpp#L642 for detailed information.

vitaut:
Are you talking about `float` because `double` conversion already uses SIMD?

xjb714:
Yes, you can refer to my implementation. However, for double, I used avx512. Your code does not involve avx512.

vitaut:
A PR to use SIMD for `float` would be welcome but I don't think we need to track this as an issue.

vitaut:
On a second thought, let's track this.

TobiSchluter:
I played around with this a little, and it seems that the right thing for `float` on x86-64 is just to modify to_bcd8 to call `get_double_significand_bcd_unshuffled_sse` followed by the unshuffling (reading this I realize that the naming is confusing `get_double_significand_bcd_unshuffled_sse` returns a shuffled string which one has to call `_mm_shuffle_XX` on to unshuffle).

Observations:
1. when using 64 bit SIMD the compilers actually substitute 128bit SIMD (with default flags, this is configurable)
2. experimenting with bswap64 instead of the full 128bit shuffle showed no performance advantages
3. it gains almost a nanosecond, i.e. ~12% (8.5ns -> 7.5ns on my AMD Ryzen)

Since I now have access to an ARM64 based mac I'll try to play the same game with the NEON code, where I would have to split the to_digits function.

It is tempting to add a to_bcd2x8 function which can also be used in the double code, but I'm mainly thinking about the case where one converts multiple numbers. The binary -> decimal conversion would still have to happen two times, but the time spent converting to a digit string would be cut in half. Maybe something for fmtlib? I mentioned before that this can be done for doubles with AVX256, but here it would be practically free in terms of code.

Edit: I just learned about the `movbe` isntruction (which is a x86-64-v3 instruction that does byte-swapping while writing to memory). That may be a better fit for writing the string to memory, so observation 2 may have to be updated for that architecture level.

vitaut:
Thanks for investigating!

> it seems that the right thing for float on x86-64 is just to modify to_bcd8 to call get_double_significand_bcd_unshuffled_sse followed by the unshuffling

Makes sense, a PR is welcome.

[...]

PR #120, opened 2026-04-10
SIMD for floats

TobiSchluter:
This adds a SIMD implementation of `to_bcd8`, i.e. float->bcd conversion, addressing #99 

In order to share code with the double implementation, this splits the NEON and x64 SIMD codes into the parts that deal with 8-digit numbers (shared between `float` and `double`) and the part that deals with 16 digits. I also changed `to_bcd8`'s interface to return the length as it is significantly faster if the length in the SSE4.1 version is evaluated before `bswap`, likely because of the shorter dependency chain this creates.

Timings are taken with @Antares0982's benchmark which measures numbers which are output in fixed format, which means a significant amount of postprocessing to place the decimal point, and it can thus show an absolute speed increase, but yhe relative one has to be taken with a grain of salt.

On my M5 notebook it goes from 5.1ns/float to 4.7. On my AMD Ryzen from 8.9ns/float to 8.3 (SSE4.1) or 8.5 (SSE2).

There are some unexplored optimization opportunities, in no particular order:
- use 64bit SIMD on Neon (would save filling the upper 64bit lane, and maybe some energy)
- add '0' inside the SIMD path, this would maybe enable writing to memory directly from the SIMD path
- if we add the '0' and write to memory, the SSE shuffle is more attractive than bswap, and could be explored
- instead of the magic algebra in the SSE2 path to fill 64bit lanes in the correct order, fill in 32bit lanes with the simpler { x / 10k, x - x / 10k, 0, 0 }
- optimize the fixed path to insert the point in SSE (BTW I tried separating the first two digits instead of only the first before doing the SIMD thing in order to create space for the point: it was faster, but that may be due to wrong algebra because in my quick attempt I couldn't make it work correctly)

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -709,22 +709,41 @@ ZMIJ_INLINE auto to_unshuffled_digits(uint32_t bbccddee, uint32_t ffgghhii,
 #endif  // ZMIJ_USE_SSE
 
 #if ZMIJ_USE_NEON
+// An optimized version for NEON by Dougall Johnson.
+  
+alignas(64) static constexpr struct neon_constants {
+  static constexpr int32_t neg10k = -10000 + 0x10000;
+  
+  using int32x4 = std::conditional_t<ZMIJ_MSC_VER != 0, int32_t[4], int32x4_t>;
+  using int16x8 = std::conditional_t<ZMIJ_MSC_VER != 0, int16_t[8], int16x8_t>;
+
+  uint64_t mul_const = 0xabcc77118461cefd;
+  uint64_t hundred_million = 100000000;
+  int32x4 multipliers32 = {div10k_sig, neg10k, div100_sig << 12, neg100};
+  int16x8 multipliers16 = {0xce0, neg10};
+} neon_consts;
+
+ZMIJ_INLINE auto to_digits_4x4digits(int32x4_t ddee_bbcc_hhii_ffgg, const neon_constants* c) noexcept
+  -> uint8x16_t {
+  // Compiler barrier, or clang breaks the subsequent MLA into UADDW + MUL.
+  ZMIJ_ASM(("" : "+w"(ddee_bbcc_hhii_ffgg)));
+
+  int32x4_t dd_bb_hh_ff =
+      vqdmulhq_n_s32(ddee_bbcc_hhii_ffgg, c->multipliers32[2]);
+  int16x8_t ee_dd_cc_bb_ii_hh_gg_ff = vreinterpretq_s16_s32(
+      vmlaq_n_s32(ddee_bbcc_hhii_ffgg, dd_bb_hh_ff, c->multipliers32[3]));
+  int16x8_t high_10s =
+      vqdmulhq_n_s16(ee_dd_cc_bb_ii_hh_gg_ff, c->multipliers16[0]);
+  return vreinterpretq_u8_s16(
+      vmlaq_n_s16(ee_dd_cc_bb_ii_hh_gg_ff, high_10s, c->multipliers16[1]));
+}
+
 // When reverse_hi_lo is true, the two 8-digit halves are returned in reversed
 // order so that trailing zeros become leading zero bytes, letting ctz count
 // them while the shuffle runs in parallel.
 template <bool reverse_hi_lo = false>
 ZMIJ_INLINE auto to_unshuffled_digits(uint64_t value) -> uint8x16_t {
-  // An optimized version for NEON by Dougall Johnson.
-  using int32x4 = std::conditional_t<ZMIJ_MSC_VER != 0, int32_t[4], int32x4_t>;
-  using int16x8 = std::conditional_t<ZMIJ_MSC_VER != 0, int16_t[8], int16x8_t>;
-  constexpr int32_t neg10k = -10000 + 0x10000;
-  alignas(64) static constexpr struct {
-    uint64_t mul_const = 0xabcc77118461cefd;
-    uint64_t hundred_million = 100000000;
-    int32x4 multipliers32 = {div10k_sig, neg10k, div100_sig << 12, neg100};
-    int16x8 multipliers16 = {0xce0, neg10};
-  } consts;
-  const auto* c = &consts;
+  const auto* c = &neon_consts;
 
   // Compiler barrier, or clang doesn't load from memory and generates 15 more
   // instructions.
@@ -744,7 +763,8 @@ ZMIJ_INLINE auto to_unshuffled_digits(uint64_t value) -> uint8x16_t {
   uint64x1_t ffgghhii_bbccddee_64 = {
       reverse_hi_lo ? (uint64_t(abbccddee) << 32) | ffgghhii
                     : (uint64_t(ffgghhii) << 32) | abbccddee};
-  int32x2_t bbccddee_ffgghhii = vreinterpret_s32_u64(ffgghhii_bbccddee_64);
+
+                      int32x2_t bbccddee_ffgghhii = vreinterpret_s32_u64(ffgghhii_bbccddee_64);
 
   int32x2_t bbcc_ffgg = vreinterpret_s32_u32(
       vshr_n_u32(vreinterpret_u32_s32(
@@ -756,17 +776,7 @@ ZMIJ_INLINE auto to_unshuffled_digits(uint64_t value) -> uint8x16_t {
   int32x4_t ddee_bbcc_hhii_ffgg = vreinterpretq_s32_u32(
       vshll_n_u16(vreinterpret_u16_s32(ddee_bbcc_hhii_ffgg_32), 0));
 
-  // Compiler barrier, or clang breaks the subsequent MLA into UADDW + MUL.
-  ZMIJ_ASM(("" : "+w"(ddee_bbcc_hhii_ffgg)));
-
-  int32x4_t dd_bb_hh_ff =
-      vqdmulhq_n_s32(ddee_bbcc_hhii_ffgg, c->multipliers32[2]);
-  int16x8_t ee_dd_cc_bb_ii_hh_gg_ff = vreinterpretq_s16_s32(
-      vmlaq_n_s32(ddee_bbcc_hhii_ffgg, dd_bb_hh_ff, c->multipliers32[3]));
-  int16x8_t high_10s =
-      vqdmulhq_n_s16(ee_dd_cc_bb_ii_hh_gg_ff, c->multipliers16[0]);
-  return vreinterpretq_u8_s16(
-      vmlaq_n_s16(ee_dd_cc_bb_ii_hh_gg_ff, high_10s, c->multipliers16[1]));
+  return to_digits_4x4digits(ddee_bbcc_hhii_ffgg, c);
 }
 #endif
 
@@ -787,7 +797,7 @@ template <> struct dec_digits<64> {
 };
 
 auto to_bcd8(uint32_t abcdefgh) noexcept -> uint64_t {
-#if !ZMIJ_USE_SSE
+#if !ZMIJ_USE_SSE && !ZMIJ_USE_NEON
   // An optimization from Xiang JunBo.
   // Three steps BCD. Base 10000 -> base 100 -> base 10.
   // div and mod are evaluated simultaneously as, e.g.
@@ -805,16 +815,28 @@ auto to_bcd8(uint32_t abcdefgh) noexcept -> uint64_t {
       neg10 * (((ab_cd_ef_gh * div10_sig) >> div10_exp) & 0xf000f000f000f);
   return is_big_endian ? a_b_c_d_e_f_g_h : bswap64(a_b_c_d_e_f_g_h);
 #else
+# if ZMIJ_USE_SSE
   const auto* c = &sse_consts;
   ZMIJ_ASM(("" : "+r"(c)));  // Load constants from memory.
 
   uint64_t abcd_efgh =
       abcdefgh + neg10k * ((uint64_t(abcdefgh) * div10k_sig) >> div10k_exp);
   auto unshuffled_bcd = _mm_cvtsi128_si64(to_digits_4x4digits(_mm_set_epi64x(0, abcd_efgh), *c));
-#  if ZMIJ_USE_SSE4_1
+#   if ZMIJ_USE_SSE4_1
   return bswap64(unshuffled_bcd);
-#  else
+#    else
   return unshuffled_bcd;
+#    endif
+#  else // ZMIJ_USE_NEON
+  const auto* c = &neon_consts;
+  ZMIJ_ASM(("" : "+r"(c)));  // Load constants from memory.
+
+  uint64_t abcd_efgh_64 =
+      abcdefgh + neg10k * ((uint64_t(abcdefgh) * div10k_sig) >> div10k_exp);
+  int32x4_t abcd_efgh = vcombine_s32(vreinterpret_s32_u64(vcreate_u64(abcd_efgh_64)), vdup_n_s32(0));
+  uint8x16_t digits_128 = to_digits_4x4digits(abcd_efgh, c);
+  return vget_lane_u64(
+    vreinterpret_u64_u8(vrev64_u8(vget_low_u8(digits_128))), 0);
 #  endif
 
 #endif  // ZMIJ_USE_SSE
--- a/zmij.cc
+++ b/zmij.cc
@@ -822,26 +822,25 @@ auto to_bcd8(uint32_t abcdefgh) noexcept -> bcd_result {
       neg10 * (((ab_cd_ef_gh * div10_sig) >> div10_exp) & 0xf000f000f000f);
   uint64_t result = is_big_endian ? a_b_c_d_e_f_g_h : bswap64(a_b_c_d_e_f_g_h);
   return {result, count_trailing_nonzeros(result)};
-#else
-# if ZMIJ_USE_SSE
+#elif ZMIJ_USE_SSE
   const auto* c = &sse_consts;
   ZMIJ_ASM(("" : "+r"(c)));  // Load constants from memory.
 
-#   if ZMIJ_USE_SSE4_1
+# if ZMIJ_USE_SSE4_1
   uint64_t abcd_efgh =
       abcdefgh + neg10k * ((uint64_t(abcdefgh) * div10k_sig) >> div10k_exp);
   uint64_t unshuffled_bcd = _mm_cvtsi128_si64(to_digits_4x4digits(_mm_set_epi64x(0, abcd_efgh), *c));
   int len = unshuffled_bcd ? 8 - ctz(unshuffled_bcd) / 8 : 0;
   return {bswap64(unshuffled_bcd), len};
-#    else
+#  else
   // Evaluate the 4 digit limbs and arrange them such that we get a result which
   // is in the correct order.
   uint64_t abcd_efgh =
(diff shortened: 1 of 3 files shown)


TobiSchluter:
It seems the CI failures are due to the C version not dealing with fixed format floats correctly.
/Users/runner/work/zmij/zmij/test/zmij-test.cc:273: Failure
Expected equality of these values:
  ftoa(43210.0f)
    Which is: "4.321e+04"
  "43210"
/Users/runner/work/zmij/zmij/test/zmij-test.cc:274: Failure
Expected equality of these values:
  ftoa(43210.1f)
    Which is: "4.32101e+04"
  "43210.1"
/Users/runner/work/zmij/zmij/test/zmij-test.cc:275: Failure
Expected equality of these values:
  ftoa(10000.f)
    Which is: "1e+04"
  "10000"

It's not clear to me why I don't see this failure.

TobiSchluter:
I fixed three small bugs which I missed before, and I only partly understand why
- structured bindings are C++14, my compilers accepted it without warning even though the language standard is correctly set in the project
- the SSE4.1 length evaluation rounded incorrectly and thus missed some cases
- the C version doesn't implement fixed point printing of floats, the error got lost in the test output and I will make sure to look at the exit status of all tests now

TobiSchluter:
I saw this in the assembly and couldn't resist improving it: the four constants used for the SSE float conversion were split across two cache lines. The latest patch rearranges the struct so that these four are at the top of the struct and thus in the same cache line.

xjb714:
@TobiSchluter 
For the f32 type, on the MacBook Air M5 processor, the performance of [xjb32](https://github.com/xjb714/xjb/blob/main/src/ftoa.cpp#L1212) has reached 3.24 nanoseconds.
Zmij can still be further optimized.

xjb714:
@TobiSchluter 
benchmark result on apple M5 :
image

TobiSchluter:
Thank you, there's definitely room for improvement on top of this very simple change

Antares0982:
I have confirmed that this PR provides a performance improvement. BTW some recent commits have introduced a significant performance regression, particularly the changes between `4baf80` and `7afa3c`.

Default SSE2 on x86-64
```
=== float benchmark (5000 rounds × 91932 values, 100 warmup) ===
  xjb                        459660000 calls       3345.48 ms total      7.28 ± 0.21 ns/call  (sink=4160850300)
  zmij_cpp                   459660000 calls       4351.60 ms total      9.47 ± 0.29 ns/call  (sink=4160472900)
  zmij-pr120                 459660000 calls       4128.80 ms total      8.98 ± 0.12 ns/call  (sink=4160472900)
  zmij-7afa3c                459660000 calls       4078.40 ms total      8.87 ± 0.10 ns/call  (sink=4160472900)
  zmij-4baf80                459660000 calls       3719.35 ms total      8.09 ± 0.10 ns/call  (sink=4160472900)
  zmij_c                     459660000 calls       3860.33 ms total      8.40 ± 0.76 ns/call  (sink=6036043800)
  zmij_rust                  459660000 calls       4062.48 ms total      8.84 ± 0.44 ns/call  (sink=4160850300)

  xjb                        459660000 calls       3940.58 ms total      8.57 ± 0.06 ns/call  (sink=7896789000)
  zmij_cpp                   459660000 calls       5093.37 ms total     11.08 ± 0.19 ns/call  (sink=7896585000)
  zmij-pr120                 459660000 calls       5054.28 ms total     11.00 ± 0.11 ns/call  (sink=7896585000)
  zmij-7afa3c                459660000 calls       5041.62 ms total     10.97 ± 0.18 ns/call  (sink=7896585000)
  zmij-4baf80                459660000 calls       4255.15 ms total      9.26 ± 0.27 ns/call  (sink=7896585000)
  zmij_c                     459660000 calls       5596.69 ms total     12.18 ± 0.12 ns/call  (sink=7896585000)
  zmij_rust                  459660000 calls       4897.24 ms total     10.65 ± 0.11 ns/call  (sink=7896789000)
```

Antares0982:
Apple M4:

```
=== float benchmark (5000 rounds × 91932 values, 100 warmup) ===
  xjb                        459660000 calls       1655.17 ms total      3.60 ± 0.05 ns/call  (sink=4160850300)
  zmij_cpp                   459660000 calls       2900.00 ms total      6.31 ± 0.08 ns/call  (sink=4160472900)
  zmij-pr120                 459660000 calls       2613.78 ms total      5.69 ± 0.05 ns/call  (sink=4160472900)
  zmij-7afa3c                459660000 calls       2857.37 ms total      6.22 ± 0.06 ns/call  (sink=4160472900)
  zmij-4baf80                459660000 calls       2683.46 ms total      5.84 ± 0.47 ns/call  (sink=4160472900)
  zmij_c                     459660000 calls       3099.43 ms total      6.74 ± 0.07 ns/call  (sink=6036043800)
  zmij_rust                  459660000 calls       3835.19 ms total      8.34 ± 0.07 ns/call  (sink=4160850300)

  xjb                        459660000 calls       2135.75 ms total      4.65 ± 0.04 ns/call  (sink=7896789000)
  zmij_cpp                   459660000 calls       2513.28 ms total      5.47 ± 0.06 ns/call  (sink=7896585000)
  zmij-pr120                 459660000 calls       2530.28 ms total      5.50 ± 0.08 ns/call  (sink=7896585000)
  zmij-7afa3c                459660000 calls       2532.05 ms total      5.51 ± 0.08 ns/call  (sink=7896585000)
  zmij-4baf80                459660000 calls       2342.43 ms total      5.10 ± 0.06 ns/call  (sink=7896585000)
  zmij_c                     459660000 calls       4163.10 ms total      9.06 ± 0.13 ns/call  (sink=7896585000)
  zmij_rust                  459660000 calls       3455.01 ms total      7.52 ± 0.06 ns/call  (sink=7896789000)
```

TobiSchluter:
The latest patch contains no functional changes to the code, it undoes a formatting change, improves a comment and makes the SSE4.1 length evaluation more readable (the assembly is identical).

TobiSchluter:
> I have confirmed that this PR provides a performance improvement. BTW some recent commits have introduced a significant performance regression, particularly the changes between `4baf80` and `7afa3c`.

I ran the benchmark script for this range of commits on my laptop, and then my little script https://github.com/user-attachments/files/24512258/find_culprit.py to find the worst offenders. These are the top slowdowns according to it. Obviously, this has to be taken with a large grain of salt, renaming variables is not known to affect performance, neither is changing SSE4.1 code when building in SSE2 mode.

```
Top 10 largest increases in zmij:

Commit: 2c32bb7dec2f78f457faa6ce09325916873b4e42
Increase: +1.22 ns (value: 30.27)
Log: 2c32bb7 Add windows ci, Fix node20 deprecation warning (#115)
------------------------------------------------------------
Commit: 45d165fe56d26823fcfb0eb58b4b236874b5d64c
Increase: +1.00 ns (value: 30.49)
Log: 45d165f More verification
------------------------------------------------------------
Commit: 6314dfdb4c524d9280055e476bb83bc4f674c43a
Increase: +0.94 ns (value: 30.31)
Log: 6314dfd Update diagram
------------------------------------------------------------
Commit: ece756f36dca2c0f53fb522c259f53b31bbc2241
Increase: +0.93 ns (value: 30.55)
Log: ece756f Apply an optimization from the xjb paper to to_decimal_fast.
------------------------------------------------------------
Commit: 9e0eec44a40d8e6ade1c74030c658475fb92dd8a
Increase: +0.93 ns (value: 30.08)
Log: 9e0eec4 num_fractional_bits -> extra_shift
------------------------------------------------------------
Commit: 50979717fb7dce08a1aa952392ea96f070968627
Increase: +0.43 ns (value: 29.37)
Log: 5097971 Clarify choice of num_fractional_bits
------------------------------------------------------------
Commit: 7afa3ccdbb4cf40131e6acb466484f13d7272723
Increase: +0.39 ns (value: 28.05)
Log: 7afa3cc Convert bcd -> ascii after shuffling. Gain 3% (#112)
------------------------------------------------------------
Commit: 6cfb9f6f9e94e6fc37e22e0c57b7a81a842454d9
Increase: +0.33 ns (value: 30.07)
Log: 6cfb9f6 Cleanup split
------------------------------------------------------------
Commit: bb0cbbd5f4418bfa3cb9bbb321310a45efe5a56e
Increase: +0.33 ns (value: 29.26)
Log: bb0cbbd Adapt verification
------------------------------------------------------------
Commit: 381914b3ea3b76639bb1bdcc6db1acf6ed03afc5
Increase: +0.30 ns (value: 30.06)
Log: 381914b Simplify to_decimal_fast
------------------------------------------------------------```

**Comment by Antares0982** [comment]
> > I have confirmed that this PR provides a performance improvement. BTW some recent commits have introduced a significant performance regression, particularly the changes between `4baf80` and `7afa3c`.
> 
> I ran the benchmark script for this range of commits on my laptop, and then my little script https://github.com/user-attachments/files/24512258/find_culprit.py to find the worst offenders. These are the top slowdowns according to it. Obviously, this has to be taken with a large grain of salt, renaming variables is not known to affect performance, neither is changing SSE4.1 code when building in SSE2 mode.
> 
> ```
> Commit: a6177758d852d4ed8606cd45d415bdcb8d1de26a
> Increase: +2.72 ns (value: 28.88)
> Log: a617775 Reintroduce digit override for round-to-even
> ------------------------------------------------------------
> Commit: 2c32bb7dec2f78f457faa6ce09325916873b4e42
> Increase: +1.22 ns (value: 30.27)
> Log: 2c32bb7 Add windows ci, Fix node20 deprecation warning (#115)
> ------------------------------------------------------------
> Commit: 45d165fe56d26823fcfb0eb58b4b236874b5d64c
> Increase: +1.00 ns (value: 30.49)
> Log: 45d165f More verification
> ------------------------------------------------------------
> Commit: e46905d925d56a094965f14b2f78ad32a35fb7b5
> Increase: +0.95 ns (value: 29.81)
> Log: e46905d Apply coding conventions
> ------------------------------------------------------------
> Commit: 6314dfdb4c524d9280055e476bb83bc4f674c43a
> Increase: +0.94 ns (value: 30.31)
> Log: 6314dfd Update diagram
> ------------------------------------------------------------
> Commit: ece756f36dca2c0f53fb522c259f53b31bbc2241
> Increase: +0.93 ns (value: 30.55)
> Log: ece756f Apply an optimization from the xjb paper to to_decimal_fast.
> ------------------------------------------------------------
> Commit: 9e0eec44a40d8e6ade1c74030c658475fb92dd8a
> Increase: +0.93 ns (value: 30.08)
> Log: 9e0eec4 num_fractional_bits -> extra_shift
> ------------------------------------------------------------
> Commit: 560815294e5ecaf435aabf95937dc80f46c9996f
> Increase: +0.78 ns (value: 27.30)
> Log: 5608152 scaled_half_ulp -> half_ulp
> ------------------------------------------------------------
> Commit: 50979717fb7dce08a1aa952392ea96f070968627
> Increase: +0.43 ns (value: 29.37)
> Log: 5097971 Clarify choice of num_fractional_bits
> ------------------------------------------------------------
> Commit: 7afa3ccdbb4cf40131e6acb466484f13d7272723
> Increase: +0.39 ns (value: 28.05)
> Log: 7afa3cc Convert bcd -> ascii after shuffling. Gain 3% (#112)
> ------------------------------------------------------------
> ```

I have some doubts about the correctness of this script's test results, since the commits `2c32bb` and `45d165` did not modify the source code. Commit `a61777` appears suspicious, it introduced an additional if statement.

**Comment by TobiSchluter** [comment]
Yeah, there's certainly a lot of noise going on, the laptop also wasn't completely unoccupied, also it only runs over the 48 revisions in the range that you indicated.

Below are the results created in the same way for -march=x86-64-v2. I also attach the raw results
v1: [results_arch_default.csv](https://github.com/user-attachments/files/26646911/results_arch_default.csv)
v2: [results_arch_x86-64-v2.csv](https://github.com/user-attachments/files/26646943/results_arch_x86-64-v2.csv)



```
Top 10 largest increases in zmij:

Commit: 4931877a939362ec938212071eaabd81e1205b4d
Increase: +1.21 ns (value: 31.15)
Log: 4931877 Implement split for subnormals
------------------------------------------------------------
Commit: a1e2911f94b483b7d0578b2d112a4caad738c237
Increase: +1.05 ns (value: 28.87)
Log: a1e2911 Don't skip exact cases during verification
------------------------------------------------------------
Commit: 6cfb9f6f9e94e6fc37e22e0c57b7a81a842454d9
Increase: +0.77 ns (value: 30.46)
Log: 6cfb9f6 Cleanup split
------------------------------------------------------------
Commit: 892c5195e62c8f690d23512098a577ede2d8b0ee
Increase: +0.75 ns (value: 30.22)
Log: 892c519 Add yy_double verification script
------------------------------------------------------------
Commit: 425cbf9e5529e88239cfe94dcbe25a366bf1940e
Increase: +0.66 ns (value: 32.19)
Log: 425cbf9 Fix gcc NEON issues
------------------------------------------------------------
Commit: 0d365c3a81bcf7175525c1a170cbeca5d837f51a
Increase: +0.46 ns (value: 30.06)
Log: 0d365c3 Attribute one of the optimizations
------------------------------------------------------------
Commit: 7afa3ccdbb4cf40131e6acb466484f13d7272723
Increase: +0.40 ns (value: 28.59)
Log: 7afa3cc Convert bcd -> ascii after shuffling. Gain 3% (#112)
------------------------------------------------------------
Commit: 427954527bf956aca61c67cbb3101fbd2ac18cec
Increase: +0.36 ns (value: 30.13)
Log: 4279545 Minor cleanup
------------------------------------------------------------
Commit: 9a0eb99fb9baebee54079ec25903b8ede966b092
Increase: +0.27 ns (value: 29.69)
Log: 9a0eb99 Minor cleanup
------------------------------------------------------------
Commit: 45d165fe56d26823fcfb0eb58b4b236874b5d64c
Increase: +0.23 ns (value: 29.92)
Log: 45d165f More verification
------------------------------------------------------------```

TobiSchluter:
I had to edit the last numbers because the benchmark script appends to the existing file, and so each revision was there twice. I also updated the results file. These also point to `a61777` which I believe is a correctness fix?

That said, looking at the list this is really not very reliable, too many comment changes.

Antares0982:
>  particularly the changes between `4baf80` and `7afa3c`.

`a61777` is not in this range. There should be another reason.

TobiSchluter:
Good catch, I was just looking at all revisions after the first one, not the range...

TobiSchluter:
I updated the result lists to not include the revisions past the change that you indicated. BTW this is with gcc. If I make it to the office tomorrow, I'll burn some cycles over lunch to do the same with more reliable benchmarkers because these results seem fairly bonkers.

zmij.cc
```diff
@@ -680,66 +654,92 @@ alignas(64) constexpr struct sse_constants {
   uint128 hundred = splat32(100);
   uint128 moddiv10 = splat16(10 * (1 << 8) - 1);
 #  endif  // ZMIJ_USE_SSE4_1
+  uint128 div10k = splat64(div10k_sig);
+  uint128 neg10k = splat64(::neg10k);
```
vitaut, line 664: Why move these?

zmij.cc
```diff
@@ -680,66 +654,92 @@ alignas(64) constexpr struct sse_constants {
...
 using m128ptr = const __m128i*;
 
-// SSE parallel version of to_bcd8: converts bbccddee and ffgghhii into
-// individual BCD digits in SIMD lane order (caller must shuffle).
-ZMIJ_INLINE auto to_unshuffled_digits(uint32_t bbccddee, uint32_t ffgghhii,
-                                      const sse_constants& c) noexcept
-    -> __m128i {
-  const __m128i div10k = _mm_load_si128(m128ptr(&c.div10k));
-  const __m128i neg10k = _mm_load_si128(m128ptr(&c.neg10k));
+ZMIJ_INLINE auto to_digits_4x4digits(__m128i y, const sse_constants& c) noexcept
```
vitaut, line 671: A comment would be nice since it's not obvious what 4x4digits mean.

zmij.cc
```diff
@@ -680,66 +654,92 @@ alignas(64) constexpr struct sse_constants {
...
+  using int32x4 = std::conditional_t<ZMIJ_MSC_VER != 0, int32_t[4], int32x4_t>;
+  using int16x8 = std::conditional_t<ZMIJ_MSC_VER != 0, int16_t[8], int16x8_t>;
+
+  uint64_t mul_const = 0xabcc77118461cefd;
+  uint64_t hundred_million = 100000000;
+  int32x4 multipliers32 = {div10k_sig, neg10k, div100_sig << 12, neg100};
+  int16x8 multipliers16 = {0xce0, neg10};
+} neon_consts;
+
+ZMIJ_INLINE auto to_digits_4x4digits(int32x4_t ddee_bbcc_hhii_ffgg, const neon_constants* c) noexcept
```
vitaut: Let's pass constants by reference for consistency with the SSE path.

zmij.cc
```diff
@@ -771,36 +771,82 @@ ZMIJ_INLINE auto to_unshuffled_digits(uint64_t value) -> uint8x16_t {
...
+  uint64_t ab_cd_ef_gh =
+      abcd_efgh +
+      neg100 * (((abcd_efgh * div100_sig) >> div100_exp) & 0x7f0000007f);
+  uint64_t a_b_c_d_e_f_g_h =
+      ab_cd_ef_gh +
+      neg10 * (((ab_cd_ef_gh * div10_sig) >> div10_exp) & 0xf000f000f000f);
+  uint64_t result = is_big_endian ? a_b_c_d_e_f_g_h : bswap64(a_b_c_d_e_f_g_h);
+  return {result, count_trailing_nonzeros(result)};
+#else
+# if ZMIJ_USE_SSE
```
vitaut: Let's make it `#elif` and reduce nesting.

vitaut (changes requested):
Thanks for another great PR! LGMT but please address minor inline comments.

vitaut:
The regression for fixed is likely due to the WIP splitting of last digit. It proved to be beneficial for the exponential format but fixed needs a bit more work. Right now we just recombine it with the rest which is obviously suboptimal: https://github.com/vitaut/zmij/blob/1649d1809ad708dfd3bd4bc68abd46bb429ad1e8/zmij.cc#L1206

We need to update the fixed path to handle it properly.

zmij.cc
TobiSchluter: Sure, I'm not happy with the name. I was following the precedent in `to_unshuffled_digits`, but then I'm also talking about decimal input digits and the output digits here are bcd digits, well, anyway, a comment is certainly called for. Added.

zmij.cc
TobiSchluter: Sure, done.

zmij.cc
TobiSchluter: Indeed, thanks!

zmij.cc
TobiSchluter: I added a comment that explains it. The intent is to keep the four constants used in the `float` path in the same cache line. Just something that struck my eye.

TobiSchluter:
Updated to reviewer comments, and rebased to the main branch. Thanks as always for the close look.

TobiSchluter:
Maybe not the greatest practice during review, but I fixed two small aesthetic issues that pained my eyes when I looked at the combined diff:
1. a nested `#if` was misaligned, fixed the column count
2. I had placed the `to_bcd8` function between the `dec_digits` declaration and its users, the `to_digits` functions, making the code harder to read for no reason. I restored the proper placement by moving the declarations close to `to_digits`. I guess I had intended to share the return type for a brief moment before thinking better of it.

vitaut:
Merged, thank you!

PR #122, opened 2026-04-13
Add CMake Threads

spinicist:
Hello,

zmij failed to link using GCC15.2 due to:

```
/software/system/gcc/15.2/bin/g++ -flto=auto -fno-fat-lto-objects  test/CMakeFiles/double-check.dir/double-check.cc.o -o test/double-check  test/libfmt.a && :
/tmp/k1078535/ccERVkvb.ltrans0.ltrans.o: In function `std::thread::_M_thread_deps_never_run()':
<artificial>:(.text+0x319): undefined reference to `pthread_create'
<artificial>:(.text+0x31e): undefined reference to `pthread_join'
collect2: error: ld returned 1 exit status
```

I am not entirely sure what `std::thread::_M_thread_deps_never_run()` does. It was simple enough to add the usual CMake incantations to get `-pthread` where needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,13 +3,9 @@ project(zmij CXX)
 
 set(ZMIJ_STANDARD cxx_std_14)
 
-set(THREADS_PREFER_PTHREAD_FLAG TRUE)
-find_package(Threads)
-
 add_library(zmij zmij.cc zmij.h)
 target_include_directories(zmij PUBLIC .)
 target_compile_features(zmij PRIVATE ${ZMIJ_STANDARD})
-target_link_libraries(zmij Threads::Threads)
 
 add_executable(example example.cc)
 target_link_libraries(example zmij)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,9 +3,13 @@ project(zmij CXX)
 
 set(ZMIJ_STANDARD cxx_std_14)
 
+set(THREADS_PREFER_PTHREAD_FLAG TRUE)
+find_package(Threads)
+
 add_library(zmij zmij.cc zmij.h)
 target_include_directories(zmij PUBLIC .)
 target_compile_features(zmij PRIVATE ${ZMIJ_STANDARD})
+target_link_libraries(zmij Threads::Threads)
 
 add_executable(example example.cc)
 target_link_libraries(example zmij)
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -39,9 +39,13 @@ target_sources(zmij-optimize-size-test PRIVATE pow10-test.cc)
 
 set(ZMIJ_CHECK_STANDARD cxx_std_20)
 
+
+set(THREADS_PREFER_PTHREAD_FLAG TRUE)
+find_package(Threads)
 add_library(fmt fmt/format.cc)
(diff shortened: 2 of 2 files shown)


Alcaro:
> I am not entirely sure what std::thread::_M_thread_deps_never_run() does

It explains itself well enough.
```c++
  private:
    // This adds to user code that creates std:thread objects (because
    // it is called by the template ctor below) strong references to
    // pthread_create and pthread_join, which ensures they are both
    // linked in even during static linking.  We can't depend on
    // gthread calls to bring them in, because those may use weak
    // references.
    static void
    _M_thread_deps_never_run() {
#ifdef GTHR_ACTIVE_PROXY
      reinterpret_cast<void (*)(void)>(&pthread_create)();
      reinterpret_cast<void (*)(void)>(&pthread_join)();
#endif
    }
```

CMakeLists.txt
```diff
@@ -3,9 +3,13 @@ project(zmij CXX)
 
 set(ZMIJ_STANDARD cxx_std_14)
 
+set(THREADS_PREFER_PTHREAD_FLAG TRUE)
+find_package(Threads)
+
 add_library(zmij zmij.cc zmij.h)
 target_include_directories(zmij PUBLIC .)
 target_compile_features(zmij PRIVATE ${ZMIJ_STANDARD})
+target_link_libraries(zmij Threads::Threads)
```
vitaut: This is not needed since threads are not used in the library itself. Please revert.

test/CMakeLists.txt
```diff
@@ -39,9 +39,13 @@ target_sources(zmij-optimize-size-test PRIVATE pow10-test.cc)
 
 set(ZMIJ_CHECK_STANDARD cxx_std_20)
 
+
+set(THREADS_PREFER_PTHREAD_FLAG TRUE)
```
vitaut: I don't think we need `THREADS_PREFER_PTHREAD_FLAG` (and also remove an extra empty newline while at it).

test/CMakeLists.txt
```diff
@@ -39,9 +39,13 @@ target_sources(zmij-optimize-size-test PRIVATE pow10-test.cc)
...
+find_package(Threads)
```
vitaut: Move it next to the usage below.

vitaut (changes requested):

test/CMakeLists.txt
spinicist: I don't appear to actually get `-pthread` unless I set this on GCC 15.2. My experience is CMake is unpredictable about this and adding the flag makes it a bit more predictable.

It's still given in the example code in their docs https://cmake.org/cmake/help/latest/module/FindThreads.html

vitaut: What error do you get without this flag and what is the compiler invocation command?

vitaut (changes requested):

test/CMakeLists.txt
spinicist: More apologies from me. I was clearly having a bad day yesterday.

It works without PREFER_PTHREADS. I've removed it. I think I've addressed your other comments as well. Let me know if there are any other problems.

CMakeLists.txt
spinicist: Reverted.

test/CMakeLists.txt
spinicist: Moved.

vitaut:
Merged, thanks!

PR #129, opened 2026-04-23
SSE improvements

TobiSchluter:
This patch combines a number of improvements to the SSE code, but only one has measurable performance impact.

Like for the float path, it makes sense to evaluate the output length before BSWAP in the SSE4.1 path.  It also simplifies the insertion of the guard bit.  This is a significant performance gain on my Zen4 (8.1ns -> 7.7ns). To my surprise it is performance neutral on my Tiger Lake laptop. I removed the mention of LZCNT because it is a microarchitecture v3 instruction which now only appears on the microarchitecture v1 path. This is the first patch.

The second patch just makes the code clearer IMO and is performance neutral, but since it changes the sequence of operations I kept it separate: in the SSE2 code, the digits for each 32bit limb are evaluated in the correct order. Currently we shuffle the 32bit lanes at the end to obtain the correct order, but it seems structurally cleaner to shuffle when actually preparing the 32bit lanes.

The third patch reorganizes the SSE code somewhat. Given the previous patch only one of the two alternatives is actually shuffled, and so I wanted to make the name less confusing and ended up, well, shuffling the code around somewhat. Again, performance neutral, the code performs exactly the same operations in the same order.

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -901,21 +901,25 @@ ZMIJ_INLINE auto to_digits(uint64_t value, bool extra_digit,
   const __m128i zeros = _mm_load_si128(m128ptr(&c.zeros));
   auto unshuffled_bcd = to_unshuffled_digits(abbccddee, ffgghhii, c);
 #  if ZMIJ_USE_SSE4_1
+  // The length is determined from the number of trailing zeros which are
+  // in the low bits before the bswap.
+  __m128i mask128 = _mm_cmpgt_epi8(unshuffled_bcd, _mm_setzero_si128());
+  uint64_t mask = _mm_movemask_epi8(mask128);
+  int len = 16 - ctz(mask | (1 << 16));
+
   const __m128i bswap = _mm_load_si128(m128ptr(&c.bswap));
   auto bcd = _mm_shuffle_epi8(unshuffled_bcd, bswap);  // SSSE3
-#  else
+  return {_mm_or_si128(bcd, zeros), len};
+#  else // ZMIJ_USE_SSE4_1
   auto bcd = _mm_shuffle_epi32(unshuffled_bcd, _MM_SHUFFLE(0, 1, 2, 3));
-#  endif
-
-  // Count leading zeros.
+  
+  // The length is determined from the number of trailing zeros which are
+  // in the high bits.
   __m128i mask128 = _mm_cmpgt_epi8(bcd, _mm_setzero_si128());
   uint64_t mask = _mm_movemask_epi8(mask128);
-#  if defined(__LZCNT__) && !defined(ZMIJ_NO_BUILTINS)
-  int len = 32 - _lzcnt_u32(mask);
-#  else
   int len = 63 - clz((mask << 1) | 1);
-#  endif
   return {_mm_or_si128(bcd, zeros), len};
+#  endif // !ZMIJ_USE_SSE4_1
 #endif  // ZMIJ_USE_SSE
 }
 
--- a/zmij.cc
+++ b/zmij.cc
@@ -756,6 +756,7 @@ ZMIJ_INLINE auto to_unshuffled_digits(uint64_t value, const constants& c)
 using m128ptr = const __m128i*;
 
 // Converts four numbers < 10000, one in each 32bit lane, to BCD digits.
+// Digits in each 32bit lane will be in order for SSE2, reversed for SSE4.1.
 ZMIJ_INLINE auto to_bcd_4x4(__m128i y, const constants& c) noexcept -> __m128i {
   const __m128i div100 = _mm_load_si128(m128ptr(&c.div100));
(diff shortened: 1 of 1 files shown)


TobiSchluter:
Here's an insane data point: so far, on my laptop I tested with g++ 13.3.0. There with and without the PR the benchmark comes in at around 27ns when using the SSE4.1 path. With clang 18.1.3 (Ubuntu version) OTOH I see 44ns (sic! slower than Dragonbox) without the PR and 26ns with it. The compiler seems to be making some terrible choice, though I can't see it in the assembly generated by godbolt's clang 18.1.0. There I see exactly the expected difference and everything else seems to be the same. I've tried this several times and tried to eliminate all sources of error because I couldn't and still can't believe this is real.

TobiSchluter:
These measurements are totally bonkers. For instance, if I align the output buffer to 64bit in `bench_random_digit` the pre-patch time with clang goes up to 48ns. So it gets worse when there's a guarantee that it doesn't cross a cache line, and I still have no idea what causes this extreme degradatation 🤷‍♂️ Or, in the sense of advertising this PR I should say "what causes this lucky outcome, which is entirely reproducible"  On my Zen4 (AMD Ryzen 6500) performance with the same compilers is not significantly affected byt this PR, the PR version is perhaps minimally faster.

vitaut:
> I can't see it in the assembly generated by godbolt's clang 18.1.0

Could you post disassembly before/after for `write<double>` from the binaries produced on your machine?

TobiSchluter:
The perf output speaks volumes:
With the patch we get 2.6 insn per cycle
```
./benchmark-clang-with-patch

Mean of per-digit medians:
zmij     : 26.77ns (26.39ns - 27.93ns) noisy

 Performance counter stats for './benchmark-clang-with-patch':

          11290.31 msec task-clock:u                     #    0.999 CPUs utilized             
                 0      context-switches:u               #    0.000 /sec                      
                 0      cpu-migrations:u                 #    0.000 /sec                      
              3453      page-faults:u                    #  305.838 /sec                      
       24272869425      cycles:u                         #    2.150 GHz                       
       63352357130      instructions:u                   #    2.61  insn per cycle            
        3828479984      branches:u                       #  339.094 M/sec                     
          26500292      branch-misses:u                  #    0.69% of all branches           
                        TopdownL1                 #      0.1 %  tma_backend_bound      
                                                  #     93.9 %  tma_bad_speculation    
                                                  #      0.1 %  tma_frontend_bound     
                                                  #      5.9 %  tma_retiring           

      11.297985873 seconds time elapsed

      11.277178000 seconds user
       0.016115000 seconds sys

```

but without the patch only 1.6!
```
./benchmark-clang-without-patch
Mean of per-digit medians:
zmij     : 44.64ns (44.22ns - 45.46ns) noisy

 Performance counter stats for './benchmark-clang-without-patch':

          18009.88 msec task-clock:u                     #    1.000 CPUs utilized             
                 0      context-switches:u               #    0.000 /sec                      
                 0      cpu-migrations:u                 #    0.000 /sec                      
              3454      page-faults:u                    #  191.784 /sec                      
       38846989536      cycles:u                         #    2.157 GHz                       
       63364833479      instructions:u                   #    1.63  insn per cycle            
        3828482478      branches:u                       #  212.577 M/sec                     
          26870978      branch-misses:u                  #    0.70% of all branches           
                        TopdownL1                 #      0.1 %  tma_backend_bound      
                                                  #     94.2 %  tma_bad_speculation    
                                                  #      1.0 %  tma_frontend_bound     
                                                  #      4.7 %  tma_retiring           

      18.012118108 seconds time elapsed

      17.990412000 seconds user
       0.019998000 seconds sys
```

Here's the expected different section in the assembly, as shown by `perf annotate`. The different part stretches from the highlighted line to the `por` a few lines down. Note that I'm on WSL so the counters aren't very reliable. I think it's worth noting for what comes below that the version with the patch clobbers `rdx` and `rax` whereas the version without the patch only touches `rax`
image
There are more differences further down in the assembly which almost all seem to be due to different choices of the register allocator, but there is one difference where the code is different.  This appears in the part right before the exponent string is inserted, and it contains the write of the digits to memory (green is with patch, the `jmp` near the top is where the fixed path jumps to the epilogue):
image
According to perf, this section isn't particularly hot.


The final length evaluation appears hot (left is with patch)
image
... but that is probably because it needs to be unified with the fixed path after speculative execution

If there were any obvious problem in the original assembly, it would also be slower on my Zen4. I'm therefore not surprised that I don't see anything obvious.

The two assemblies are here:
[asm-without-patch.txt](https://github.com/user-attachments/files/27085116/asm-without-patch.txt)
[asm-with-patch.txt](https://github.com/user-attachments/files/27085117/asm-with-patch.txt)

vitaut:
Merged, thanks!

PR #135, opened 2026-04-30
Correct fix: reenable SSE on 32-bit MSVC builds (fixes vitaut#132): Added extract_u64.

IRainman:
I do this because support for Windows 10 still isn't ended, and Firefox 115 is still supported on all 32-bit Windows from Windows 7 to Windows 10.


vitaut:
Thank you

Issue #40, opened 2025-12-28
Integers?


AlexGuteniev:
The `to_bcd8` should beat an usual `itoa` implementation.

Wanna add integers API?

TobiSchluter:
I was thinking the same thing when I was working on it just now.

A 32bit int can have at most 10 digits,  a 64bit int 19 digits, so it doesn't map cleanly to what we have right now, though. Also, integers will have leading zeros whereas our significands have at most one, and cutting off the high digits would require one copy if using `to_bcd8` (or maybe a bit-count followed by a shift, if it fits in a register)

dtolnay:
This is the best variation I have found so far: https://github.com/dtolnay/itoa/pull/66. 4-digit bcd using 32-bit multiplication (as opposed to bcd8 with 64-bit multiplications) can be done four at a time using u32x4 simd. This ends up being significantly faster than the best implementation I've found involving bcd8. Both are implemented in https://github.com/dtolnay/itoa-benchmark, you can reproduce this result by running `cargo run --release bcd4:u64 bcd8:u64`.

vitaut:
This is a bit out of scope but I won't object to a PR.

TobiSchluter:
@dtolnay is your benchmark repeating the same numbers of digits? Then you will get perfect branch prediction on the choice between the different BCD versions. This may or may not be realistic for your application. 

I would think that indeed large integers are much rarer than short ones, so a realistic sampling would probably need to use a power law distribution or something.

dtolnay:
The chart in the readme is done using a similar methodology as <https://github.com/miloyip/itoa-benchmark>'s "Random" benchmark, which is to measure 100k values chosen randomly from {0, 1, 2, …, 9}, then a separate measurement for 100k values chosen randomly from {10, 11, …, 99}, etc through {1000000000, …, 4294967295}. So within each measurement the length is perfectly predicted.

Here is a different methodology that makes the length unpredictable: https://github.com/dtolnay/itoa-benchmark/pull/17. It generates 100k values of mixed length. For uint32 this is 10k values of each length ranging from 1 through 10. The baseline is how long it takes to format these 100k values shuffled in random order. Then separately it measures formatting the same 100k values of mixed length plus another 100k values of a specific length, again all shuffled together. The measurement reported in the chart is the difference of this from baseline.

[...]

PR #137, opened 2026-05-07
Shuffle less for SSE4.1

TobiSchluter:
In the current code, we shuffle to reverse the byte order and we shuffle to shift the bits. This combines the two. It also replaces a slow `memmove` for the decimal digits with another shuffle + store, so that the data is never re-read from memory.

This is my first AI-assisted patch to zmij, and the AI's main task was to figure out the special cases when moving the writing of the last digit behind the shuffles. Previously, the last_digit only sometimes was part of the memmove and could thus end up in  a different position. The AI suggested to keep a table of where it goes, and so I did that. A table-free approach made the compiler add a branch and lose time.

The way I avoided to increase the table size for the various bswap + shift variations and thus messing up alignments or introducing padding is quite simple: what we want to do is to reverse the order and drop the first N digits, where N depends on whether its the integral part or the decimals, and it depends on the exponent. We don't care what is added in the end past the actual digits. So, it doesn't actually matter what the last few entries in the shuffler are, and so we just keep the same bswap shuffle table, read it at an offset and don't care what the entries at the end end up being. This only works because we reverse at the same time, so this cannot be carried to the NEON code.  The additional bytes are loaded from the same struct, so it's not uninitialized memory and ASAN is happy, but I'm not sure what the standard says about this. I don't think the SIMD memory access is in its scope. BTW I changed the type of `bswap` to `char *` in order to get simpler index algebra and less casts.

Finally, in order to keep everything branch-free, in the case of `dec_exp < 0` where no decimal point is inserted between the digits, the identical shuffle + store operation is simply repeated. This turned out to be measurably faster than conditionally jumping past it. Note that the zmij/canada benchmark, given Canada's geography, doesn't contain any numbers < 1, so I created a separate benchmark (not included) that uses random fixed numbers. To my surprise it was actually faster than the processing of the Canada data. A possible explanation is that it triggered the `last_digit` far less often but I didn't check this in detail.

On my Zen 4 this saves 5%, on my Tiger Lake it is performance neutral.

I also asked Claude to do try the same thing for NEON but none of the variations it tried came out faster than what is there right now.

ps I just saw that I didn't clean this up fully. Hence the non-squashed three-patch series with one unrelated change that I didn't mean to commit as part of this: I moved the `alignas(64)` annotation from the `static_data` declaration to the `struct data` declaration. The idea being that this way all the pointers that we use remember the alignment. Turns out the assembly with and without this change is the same on x64 https://godbolt.org/z/Mxz1fbKjY and ARM64 https://godbolt.org/z/ecnPzTWhK, but I think it's cleaner, so I'm not going to remove it right now even though it is unrelated.

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -1300,11 +1300,11 @@ auto write(Float value, char* buffer) noexcept -> char* {
 
   // Write significand/fixed.
   char* start = buffer;
-  auto dig = to_digits<traits::num_bits>(dec.sig, *d);
   constexpr int bcd_size = traits::num_bits == 64 ? 16 : 8;
   if (dec_exp >= traits::min_fixed_dec_exp &&
       dec_exp <= traits::max_fixed_dec_exp) {
     memcpy(start, &zeros, 8);  // For dec_exp < 0.
+    auto dig = to_digits<traits::num_bits>(dec.sig, *d);
     char last_digit = '0' + (-has_last_digit & dec.last_digit);
     int num_digits = select(has_last_digit, bcd_size, dig.num_digits - 1);
 
@@ -1317,18 +1317,18 @@ auto write(Float value, char* buffer) noexcept -> char* {
     buffer += layout.start_pos;
 #if ZMIJ_USE_SSE4_1
     if (bcd_size == 16) {
-      // dig.digits is uint64_t for float, alias as __m128i to avoid a compiler error.
-      auto& digits = reinterpret_cast<const __m128i&>(dig.digits);
       // Assemble the digit string (minus the last_digit and, for has_extra_digit
       // == 1, BCD[15]) in a single SIMD register and store it in one go.
       __m128i tbl = _mm_load_si128(m128ptr(&layout.shuffle[has_extra_digit]));
-      __m128i out = _mm_shuffle_epi8(digits, tbl);
-      memcpy(buffer, &out, bcd_size);
+      // dig.digits is uint64_t for float, alias as __m128i to avoid a compiler error.
+      auto& digits = reinterpret_cast<const __m128i&>(dig.digits);
       // For has_extra_digit == 1 with 0 <= dec_exp <= 14 the BCD[15] byte falls
       // outside the vector at buffer[16]; write it unconditionally.  In the
       // other cases (dec_exp == 15, dec_exp < 0, or has_extra_digit == 0) it's
       // either already in the vector or overwritten below by '.' / last_digit.
       buffer[bcd_size] = (char)_mm_extract_epi8(digits, 15);
+      __m128i out = _mm_shuffle_epi8(digits, tbl);
+      memcpy(buffer, &out, bcd_size);
       start[layout.point_pos] = '.';
       buffer[layout.last_digit_pos[has_extra_digit]] = last_digit;
       return buffer + layout.end_pos[num_digits + has_extra_digit - 1];
@@ -1341,6 +1341,7 @@ auto write(Float value, char* buffer) noexcept -> char* {
     start[point_pos] = '.';
     return buffer + layout.end_pos[num_digits + has_extra_digit - 1];
   }
+  auto dig = to_digits<traits::num_bits>(dec.sig, *d);
   if (traits::num_bits == 32 && exp_float_shuffle_table::enable) {
     uint64_t exp_data = d->exp_strings.data[dec_exp + exp_string_table::offset];
     return write_exp_float_simd(buffer, dig, dec.last_digit, has_last_digit,
--- a/zmij.cc
+++ b/zmij.cc
@@ -610,12 +610,30 @@ struct fixed_layout_table {
       traits::max_fixed_dec_exp - traits::min_fixed_dec_exp + 1;
 
   // On AArch64, align entry to 32 bytes so indexing uses `lsl #5` not `umaddl`.
-  struct alignas(ZMIJ_AARCH64 && !ZMIJ_OPTIMIZE_SIZE ? 32 : 1) entry {
-    // Byte offset past leading "0.00..." before first significant digit.
+  // On x86 with SSE4.1, align entry to 64 bytes (one cache line) and put the
+  // 16-byte-aligned shuffle table at the back: the scalar fields and the
+  // shuffle for any single dec_exp then share one cache-line fill.
+  struct alignas(ZMIJ_AARCH64 && !ZMIJ_OPTIMIZE_SIZE ? 32
+                 : ZMIJ_USE_SSE4_1
+                    ? (ZMIJ_OPTIMIZE_SIZE ? 16 : 64) : 1) entry {
+#if ZMIJ_USE_SSE4_1
+    // pshufb shuffle table that places BCD bytes in their final output slots,
+    // with the byte at the decimal-point position (if any) set to a pshufb
+    // "zero" marker (high bit set).  Indexed by extra_digit.
+    //  Needs 16 byte alignment because of aligned load below.
+    alignas(16) unsigned char shuffle[2][16];
+#endif
+
+// Byte offset past leading "0.00..." before first significant digit.
     unsigned char start_pos;
     unsigned char point_pos;
     // Start position for shifting digits right by one to insert the point.
     unsigned char shift_pos;
+#if ZMIJ_USE_SSE4_1
+    // Buffer-relative position of the last_digit byte, indexed by has_extra_digit.
+    // Only used for bcd_size == 16 (doubles).
+    unsigned char last_digit_pos[2];
+#endif
     // Offset past the end of fixed-notation output, indexed by sig length - 1.
     unsigned char end_pos[traits::max_digits10];
   };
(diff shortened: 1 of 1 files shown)


vitaut:
> It also replaces a slow memmove for the decimal digits with another shuffle + store, so that the data is never re-read from memory.

Which `memove` exactly are you referring to and would it make sense to split it into a separate PR to evaluate optimizations effects independently? Or is it a side-effect of the main optimization?

zmij.cc
```diff
@@ -673,8 +691,12 @@ struct data {
 #  if ZMIJ_USE_SSE4_1
   uint128 neg100 = splat32(::neg100);
   uint128 neg10 = splat16((1 << 8) - 10);
-  uint128 bswap = uint128{pack8(15, 14, 13, 12, 11, 10, 9, 8),
-                          pack8(7, 6, 5, 4, 3, 2, 1, 0)};
+  // We will read from bswap at offsets for fixed format output.  Accomodate
+  // the necessary offset calculations by using char*.
+  // This is aligned to 64bits, so we won't cross a cache line when reading,
```
vitaut: bits or bytes?

zmij.cc
```diff
@@ -673,8 +691,12 @@ struct data {
...
+  // even with offsets.
+  char bswap[16] = {15, 14, 13, 12, 11, 10, 9,  8,
```
vitaut: Do I understand correctly that we can read past the end of this array (and get part of the content of the next field)?

TobiSchluter:
> > It also replaces a slow memmove for the decimal digits with another shuffle + store, so that the data is never re-read from memory.
> 
> Which `memove` exactly are you referring to and would it make sense to split it into a separate PR to evaluate optimizations effects independently? Or is it a side-effect of the main optimization?

The second one which reads the data immediately written before. This dominated the profile.

I will commit a simplification of the evaluation of the new stuff in float_layout tonight. It's probably not surprising that the AI stuff was a bit too roundabout. I will then split this in, probably, three patches which can be benchmarked in sequence: 1. Move writing `last_digit` to the end with helper entries in `fixed_layout` 2. Merge bswap and writing of digits 3. Replace `memmove` for decimal places with another shuffle

zmij.cc
TobiSchluter: Good eyes! Will correct

TobiSchluter:
@vitaut so I made the mistake of benchmarking not only with gcc but also with clang while reworking the patch: the whole thing is a bit of a mirage, with clang this is actually a pessimization, and the speed up with gcc only brings it to up to par with the pesssmized clang. So, I think this patch is misguided, and the apparent gain is actually due to gcc being led away from some questionable choices it makes in the original version. OTOH thinking about this, it became clear that I should just do what I had considered but dismissed, namely put together the string and put the decimal point in the middle while still inside the SSE register, and only do a single store to memory (+ extra_digit and last_digit). This seems to work well, and I hope to have it ready for submission later this week.

vitaut:
Thanks for the detailed analysis. This sounds like a good plan and I look forward to the updated version.

TobiSchluter:
This takes a bit longer than planned because I went down a nano-optimization rabbit-hole on this and got distracted by the other patch I just posted. I wanted to finish this today, but due to a sick child it will have to wait until Monday.

vitaut:
No rush and hope your child gets better soon.

TobiSchluter:
I'm sorry, I had written a long comment about the changes, but seem to not have submitted it 🫤 Sorry about that, will recover it later.

TobiSchluter:
I still had the edit window open on this PC, so I could restore the message:

This series became more interesting than I'd hoped, and so it took a bit longer to converge on something. Everything I tried beyond this -- if it made any change at all -- made clang regress and gcc improve. Seems like compilers don't like producing optimal code. Here are the numbers on my Zen4:
| compiler | bench | base | test | Δ% | p | verdict |
|---|---|---|---|---|---|---|
| gcc | zmij (scientific) | 7.51ns | 7.56ns | +0.6% | 1.3e-3 | SLOWER (marginal) |
| gcc | zmij/canada | 7.50ns | 7.21ns | **−3.9%** | 4.6e-156 | FASTER |
| gcc | zmij/fixed_range | 7.51ns | 7.27ns | **−3.3%** | 6.6e-49 | FASTER |
| clang | zmij (scientific) | 6.81ns | 6.72ns | **−1.4%** | 2.5e-14 | FASTER |
| clang | zmij/canada | 7.22ns | 6.57ns | **−8.9%** | 3.6e-231 | FASTER |
| clang | zmij/fixed_range | 6.73ns | 6.56ns | **−2.5%** | 2.9e-19 | FASTER |

You will note that I added a new benchmark. This exercises the whole exponent range for the fixed numbers. Canada is situated in a geographical situation that limits the range of numbers encountered, and small numbers exercise a slightly different code path (the initial zeros are used), and so it's worth making sure that there is no regression.

You will also notice that the benchmark numbers are within the 2-3% speedup range except for clang's zmij/canada test where a massive gain obtained. This is the result of the third patch in the series, which does something very simple (sorry about the AI generated commit message, I only noticed as I'm writing this that I didn't turn it into something readable). This patch moves the to_digits call on the fixed path after the memcpy which puts the zeros in place. I thought that a `memcpy` to a `char *` might be allowed to alias something that is used and thus may pessimize the code, so I tried this change of order. I'm not sure if that's the explanation, but clang suddenly sped up a few notches, so I'm happy. Now to give you an illustration of how random this is: on my Tiger Lake laptop this patch is actually performance neutral and doing only the change of order has neglible performance impact.

What the patch does is not very complicated: instead of `memcpy`íng  the digits around to create space for the decimal point, it shuffles them around while still in a SIMD register. For that it uses a precalculated shuffle table, and in order to be able to efficiently insert the last_digit, it uses a table of insertion positions contained in an array indexed by extra_digit and the length of the significand. I also changed the length ternary to a `select()` because amid the chain of patches, gcc decided that the existence of last_digit should be predictable, and it could jump, making performance regress badly.

I will not bore you with a recount of all the things I tried, as you can tell from the impact of the reordering, it can be quite unpredictable how things will benefit.

vitaut:
Thanks for the update!

The new benchmark makes sense but please move it to a separate PR and also rebase this one since there were some changes to the shuffle table.

TobiSchluter:
Updated to current main and separated out the fixed benchmark in #143
@vitaut looking forward to your thoughts on both.

zmij.cc
TobiSchluter: Yes, the contents of the last entry don't matter (it will be the low byte of div10k_sig). I think this is pre-existing, but I'd have to dig in the history to confirm.

I'm not sure why I see this comment from May 10 only now, sorry about the late reply.

vitaut:
Confirmed the improvement on gcc on the first commit (rebased):

```
$ test/abtest.py c5998fb9a7ec37a55d931ef05138d296074962cc 67f865d878a1e307bf4d8aa09
eecac3bac2b411d
benchmark                    base       test      Δ%         95% CI        p  verdict
zmij/fixed_range          14.38ns    13.18ns   -8.3%  [ -9.4, -7.3]  2.6e-55  FASTER
zmij/canada               14.11ns    13.21ns   -6.3%  [ -7.8, -4.9]  4.8e-18  FASTER
zmij                      13.93ns    13.80ns   -0.9%  [ -2.1, +0.4]     0.17  NOISE
dragonbox/canada          22.79ns    22.63ns   -0.7%  [ -2.1, +0.7]     0.32  NOISE
dragonbox                 44.39ns    44.35ns   -0.1%  [ -0.7, +0.6]      0.8  NOISE
dragonbox/fixed_range     35.25ns    35.27ns   +0.1%  [ -0.6, +0.7]     0.88  NOISE
```

and no effect on gcc on the second commit:

```
benchmark                    base       test      Δ%         95% CI        p  verdict
zmij/fixed_range          13.20ns    13.38ns   +1.3%  [ +0.2, +2.5]    0.026  NOISE
zmij                      13.80ns    13.90ns   +0.7%  [ +0.1, +1.4]    0.031  NOISE
zmij/canada               13.31ns    13.33ns   +0.1%  [ -1.2, +1.4]     0.88  NOISE
dragonbox                 44.32ns    44.35ns   +0.1%  [ -0.3, +0.4]     0.76  NOISE
dragonbox/fixed_range     35.23ns    35.24ns   +0.0%  [ -0.4, +0.5]     0.84  NOISE
dragonbox/canada          22.52ns    22.51ns   -0.0%  [ -0.5, +0.5]      0.9  NOISE
```

zmij.cc
TobiSchluter: oh, i didn't see it because the code changed and no longer used this. And I was staring at old code and commenting based on that. So I deleted some noise in this thread. Sorry about the confusion. I still don't understand why github sent me a mail pointing to this comment.

vitaut: No worries, I think I also forgot to publish this comment during the initial review so it showed up much later.

vitaut:
The results on my clang version on epyc are mixed though (d1d9c2daf9811009fd6154a53436d1bf0042b3b6 vs 686bfa014d124264a2040884ef398d8f2a01514d):

| benchmark         | base   | test   | typical Δ%   | consistent verdict   |
|-------------------|--------|--------|--------------|----------------------|
| zmij/canada       | ~16.7ns | ~16.4ns | −2% to −1%  | FASTER (once NOISE)  |
| zmij              | ~21.0ns | ~21.2ns | +1%         | SLOWER (once NOISE)  |

So I've merged the first commit only for now.

vitaut:
I'm going to close this PR for now and for clang changes I suggest opening a new PR with a bit more detail about the improvements and how to repro them. Thanks!

TobiSchluter:
Thanks! Also for looking at the diffs separately.

I've found it remarkably hard to get reliable speed measurements. Not only do different microarchitectures have different eprformance characteristics, but there is also less predictable stuff. E.g., with clang19 there's sometimes a massive speed regression on zen5 that I cannot at all understand from the assembly. Opus is hypothesizing that it's about the code being so small that more than one iteration of the benchmark fits in the out-of-order buffer, and one thus sees unrealistic and unpredictable pressure on the multiplication and shuffle ports. I've no idea if that holds, it also asserts a lot of things that are easy to disprove, but given the things that I saw with integer conversion #40 I'm willing to assume that CPUs indeed look at a lot of code simultaneously.

Anyhow, I will make sure to include more details of the benchmarks going forward.

Issue #133, opened 2026-04-30
Optimize float handling more


vitaut:
`float` performance has been improved and we now have a benchmark. However, there is still a significant perf gap that should be closed:

ftoa-benchmark:
```
---------------------------------------------------------------------------
Benchmark           Time             CPU   Iterations Throughput Time/float
---------------------------------------------------------------------------
dragonbox    17521941 ns     17499585 ns           41   51.43M/s    19.44ns
zmij          6724118 ns      6712660 ns          103  134.08M/s     7.46ns
```

dtoa-benchmark:
```
----------------------------------------------------------------------------------
Benchmark                 Time             CPU   Iterations Throughput Time/double
----------------------------------------------------------------------------------
dragonbox          44899075 ns     44879062 ns           16   37.88M/s     26.40ns
dragonbox/canada    1518293 ns      1517635 ns          463   73.22M/s     13.66ns
zmij               11016342 ns     11009714 ns           63  154.41M/s      6.48ns
zmij/canada          740330 ns       739017 ns          955  150.37M/s      6.65ns
```

vitaut:
Investigated this on current `main` (`c864560f`) using an Apple M5 Max with Apple Clang 21, Release mode and SIMD enabled.
The mixed-workload gap still reproduces:
- `float`: 4.86 ns/value
- `double`: 3.99 ns/value

However, the conversion core does not appear to be the cause. The benchmark generates independent corpora - nine digit groups for `float`, seventeen for `double` - then shuffles them. Their notation distributions are:
- `float`: 129,545 / 900,000 fixed (14.394%)
- `double`: 55,443 / 1,700,000 fixed (3.261%)

This matters because `write` branches between fixed and scientific output. Grouping the exact same values by notation, without changing the corpus, produced:
- `float`: 4.75 ns mixed → 3.61 ns grouped
- `double`: 3.91 ns mixed → 3.72 ns grouped

Direct `to_decimal` measurements were approximately 2.09 ns for both types.

So this is mostly a benchmark artifact, not a meaningful perf difference.

PR #140, opened 2026-05-15
Build float string inside SIMD register, gain 30%

TobiSchluter:
This patch uses a table-driven approach to build the whole float string inside a SIMD register when outputting scientific notation. With clang 13.3 on Ubuntu this gains the advertised 30%, but that is because with this compiler, the pre-patch version is really slow.  Let me begin with the timings on my Tiger Lake:

clang 18.1.3:
| benchmark     |   base  |     test  |    Δ%   |      95% CI    |    p  verdict|
|--------|--------|--------|--------|--------|--------|
|zmij         | 18.18ns |   12.38ns | -31.9% | [-33.8,-30.0] |  2e-237  FASTER |
|dragonbox |    21.00ns |   20.51ns  | -2.3% | [ -4.3, -0.3] |   0.022  NOISE |

gcc 13.3.0:
| benchmark     |   base  |     test  |    Δ%   |      95% CI    |    p  verdict|
|--------|--------|--------|--------|--------|--------|
|zmij     |     13.67ns  |  13.15ns |  -3.8% | [ -5.9, -1.8] | 0.00019  FASTER |
|dragonbox  |   24.71ns |   24.54ns |  -0.7% | [ -2.5, +1.1] |    0.46  NOISE |

On Zen4 the gains are 5.9% and 7.7% respectively with more recent compilers, on M5 Mac it's on 0.7% which may not be enough to justify adding the tables (544 bytes).

So, what do the patch do? A float number is short enough to fit into an 16byte register together with the exponential string. We build a shuffle table that tells us where each digit from the byte-reversed BCD conversion, the decimal point and the exponent goes. The table can be seen in its entirety here https://godbolt.org/z/qbvP8TPMx Typical entries look like this:
```
ndigits 6
  has_extra_digit 0 has_last 0 mask [6, 13, 5, 4, 3, 2, 8, 9, 10, 11, 128, 128, 128, 128, 128, 128]
  has_extra_digit 0 has_last 1 mask [6, 13, 5, 4, 3, 2, 1, 0, 12, 8, 9, 10, 11, 128, 128, 128]
  has_extra_digit 1 has_last 0 mask [7, 13, 6, 5, 4, 3, 2, 8, 9, 10, 11, 128, 128, 128, 128, 128]
  has_extra_digit 1 has_last 1 mask [7, 13, 6, 5, 4, 3, 2, 1, 0, 12, 8, 9, 10, 11, 128, 128]
```
We arrange the input such that the converted digits are in bytes 0...7; 8, 9, 10, 11 are the exponent e+00, loaded from the other table, byte 13 is '.' and 12 is the extra digit.

The table is indexed by nDigits, has_extra_digit and has_last_digit, but makign it a flat array and calculating the offset directly turned out the fastest. Same goes for the order of indices which allows a two-step LEA evaluation, and only the second step (ndigits) doesn't proceed in parallel with the bcd. The order of the entries in the SIMD register is also the result of careful optimization to minimize the number of operations.

Finally, I tried evaluating the exponent BCD together with the significand to avoid the exp_table, but that added the insertion into the SIMD register on the top of the dependency chain whereas the table look up in the current version can be evaluated in parallel with the BCD conversion, which reduces the cost to zero. I investigated also prefixing the sign in the same way, but that would either double the size of the table or add another shuffle for no gain as long as the write-out buffer is not guaranteed to be aligned. With a gigantic table or a split shuffle lookup one could also do the fixed-formatting with the same code, but on a quick attempt I couldn't make it faster. It would need to tabulate a start position and where to put the decimal point as a function of the decimal exponent.

This partly addresses issue #133. Maybe someone who is more versed in NEON SIMD can also improve the timings that you gave in that PR :)

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -540,6 +540,79 @@ struct exp_string_table {
   }
 };
 
+// Shuffle masks for branchless float scientific-notation output. A single
+// pshufb assembles "<d>.<rest><e±xx>" from a source register holding the
+// unshuffled (LSD-first) ASCII digits in bytes 0..7, the exp string
+// in bytes 8..11, the decimal point (byte 12 = '.'), and the last_digit
+// (ASCII, byte 13). The shuffle indices reverse the byte order while
+// inserting the decimal point and placing the exponent after the last
+// non-zero digit.
+//
+// In unshuffled form: for extra_digit, MSD lives at byte 7 (LSD at 0);
+// for the leading-zero-padded case (no extra_digit), MSD is at byte 6
+// and byte 7 holds the padding '0'.
+//
+// After benchmnarking, the fastest way of indexing appears to be a flat
+// array with index = (num_digits - 1) * 4 + has_last_digit * 2 + extra_digit.
+struct scientific_float_mask_table {
+  static constexpr bool enable = ZMIJ_USE_SSE4_1 || ZMIJ_USE_NEON;
+
+  alignas(16) unsigned char masks[enable ? 32 * 16 : 1] = {};
+  unsigned char lengths[enable ? 32 : 1] = {};
+
+  constexpr auto get_index(int ndigits, int has_last, int has_extra_digit) const noexcept {
+    return (ndigits - 1) * 4 + has_last * 2 + has_extra_digit;
+  }
+
+  struct entry {
+    const unsigned char* mask;
+    unsigned char length;
+  };
+
+  constexpr auto get_entry(int ndigits, bool has_last, bool has_extra_digit) const noexcept {
+    auto idx = get_index(ndigits, has_last, has_extra_digit);
+    return entry{ &masks[idx * 16], lengths[idx] };
+  }
+
+  constexpr scientific_float_mask_table() {
+    if (!enable) return;
+    for (int ndigits = 1; ndigits <= 8; ++ndigits) {
+      for (int has_last = 0; has_last < 2; ++has_last) {
+        for (int extra_digit = 0; extra_digit < 2; ++extra_digit) {
+          int idx =  get_index(ndigits, has_last, extra_digit);
+          unsigned char* out = &masks[idx * 16];
+          for (int i = 0; i < 16; ++i) out[i] = 0x80;  // zero by default
+
+          // Position of the MSD in the LSD-first unshuffled byte order.
+          unsigned char msd_byte = extra_digit ? 7 : 6;
+          unsigned char length = 0;
+          if (has_last) {
+            // Always 8 BCD chars in the significand plus a last-digit char;
+            // for !extra_digit the leading '0' of the 8-digit padded BCD is shown.
+            out[length++] = msd_byte;
+            out[length++] = 13;  // '.'
+            for (int i = 0; i < (extra_digit ? 7 : 6); ++i)
+              out[length++] = msd_byte - 1 - i;
+            out[length++] = 12;  // last_digit
+          } else {
+            length = (extra_digit + ndigits == 2) ? 1 : extra_digit + ndigits;
+            out[0] = msd_byte;
+            if (length > 1) {
+              out[1] = 13;  // '.'
+              for (int i = 0; i < length - 2; ++i)
+                out[2 + i] = msd_byte - 1 - i;
+            }
+          }
+          // Exp string follows the significand.
+          for (unsigned char i = 0; i < 4; ++i)
+            out[length++] = 8 + i;
+          lengths[idx] = length;
+        }
+      }
+    }
+  }
+};
+
 // Per-decimal-exponent buffer layout for branchless fixed-notation output.
 // Each entry holds the byte positions of the leading zeros, decimal point,
 // and end of output, indexed by the decimal exponent (dec_exp).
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR!

I can see ~10% speedup on an unspecified EPYC Genoa:

```
$ test/abtest.py --float 6c8fcbc9e1dde4f52f1d8c7a77914b22f11983e7 a320c264d3293c004470afaae5b204de95aa5131
...
benchmark        base       test      Δ%         95% CI        p  verdict
zmij          15.32ns    13.78ns  -10.1%  [-10.5, -9.6]        0  FASTER
dragonbox     31.90ns    31.91ns   +0.0%  [ -0.2, +0.2]     0.74  NOISE
```

and no change on EPYC 7C13:

```
$ test/abtest.py --float 6c8fcbc9e1dde4f52f1d8c7a77914b22f11983e7 a320c264d3293c004470afaae5b204de95aa5131
...
benchmark        base       test      Δ%         95% CI        p  verdict
zmij          13.61ns    13.64ns   +0.3%  [ -0.1, +0.6]     0.13  NOISE
dragonbox     23.72ns    23.75ns   +0.1%  [ +0.0, +0.2]  0.00088  SLOWER
```

zmij.cc
```diff
@@ -839,6 +914,23 @@ template <int num_bits> struct dec_digits {
...
+  __m128i unshuffled;
+  int num_digits;
+};
+#elif ZMIJ_USE_NEON
+template <> struct dec_digits<32> {
+  uint64_t digits;
+  uint8x16_t unshuffled;
+  int num_digits;
+};
+#endif
```
vitaut: Let's move conditional compilation into the struct similarly to the 64-bit specialization. Also the main struct can probably reduce to a declaration.

zmij.cc
```diff
@@ -938,6 +1055,56 @@ ZMIJ_INLINE void write_digits(char* buffer, uint64_t digits,
   memmove(buffer, buffer + drop_leading_zero, sizeof(digits));
 }
 
+#if ZMIJ_USE_SSE4_1
```
vitaut: Let's move conditional compilation into the function body.

zmij.cc
```diff
@@ -1149,6 +1316,15 @@ auto write(Float value, char* buffer) noexcept -> char* {
     start[point_pos] = '.';
     return buffer + layout.end_pos[num_digits + extra_digit - 1];
   }
+#if ZMIJ_USE_SSE4_1 || ZMIJ_USE_NEON
+  if (traits::num_bits == 32 && exp_string_table::enable) {
```
vitaut: Should this be guarded by `scientific_float_mask_table::enable`?

zmij.cc
```diff
@@ -938,6 +1055,56 @@ ZMIJ_INLINE void write_digits(char* buffer, uint64_t digits,
...
+    bool has_last_digit, bool extra_digit, uint64_t exp_data,
+    const data& d) noexcept -> char* {
+  uint8x16_t ascii = vorrq_u8(dig.unshuffled, vdupq_n_u8('0'));
+  uint32_t prefix = (uint32_t('.') << 8) + uint32_t('0') + last_digit_value;
+  uint64_t hi_qword = (exp_data & 0xFFFFFFFFu) | (uint64_t(prefix) << 32);
+  uint8x16_t src = vreinterpretq_u8_u64(
+      vsetq_lane_u64(hi_qword, vreinterpretq_u64_u8(ascii), 1));
+
+  auto m = d.scientific_float_masks.get_entry(dig.num_digits, has_last_digit, extra_digit);
+  int idx = (dig.num_digits - 1) * 4 + int(has_last_digit + 2 * extra_digit);
```
vitaut: idx is not used.

vitaut (changes requested):

zmij.cc
TobiSchluter: done

zmij.cc
TobiSchluter: sorry, yes, fixed. though this is overridden anyway by the merging of the two functions

zmij.cc
TobiSchluter: yes, and scientific_float_mask_table depends on exp_string_table, changed thusly

zmij.cc
TobiSchluter: done, though I kept one #ifdef around the function this definitely reduces duplication.

TobiSchluter:
BTW thinking about this a bit more, I should probably not click "resolve conversation", sorry if that meant you had to do extra work to read my – admittedly not very substantial – responses.

For the type rewrite, I considered using SFINAE to avoid the unused member in the non-SIMD case. This is really just shorthand notation that gets optimized away, so I don't think it's worth it.

TobiSchluter:
I couldn't resist an obvious improvement that I had missed before: there is no need for a separate array to keep the length of the output strings. The last byte of the mask is always beyond the end of the string, so it doesn't matter what we put there. Now we reuse it for the length. This saves an additional load which may end up loading a different cache line, and the address calculation also becomes simpler (the offset between the mask and its length is now a fixed 15, where previously it was variable).

I also improved some of the comments (the big comment was outdated and the information that was in it is repeated later anyway), and I tried to make the evaluation of the masks a little clearer.

This has no measurable performance impact, but saves 32 bytes, taking the table from 544 bytes to 512.

vitaut:
Merged, thank you!

Issue #141, opened 2026-05-24
__builtin_is_constant_evaluated doesn't exist in VS2017


archo5:
when building zmij.cc for x64 with VS2017:

```
zmij.cc(264): error C3861: '__builtin_is_constant_evaluated': identifier not found
zmij.cc(443): error C2131: expression did not evaluate to a constant
zmij.cc(264): note: a non-constant (sub-)expression was encountered
```

issue seems to have been introduced in https://github.com/vitaut/zmij/pull/48

the relevant block of code probably just needs to be wrapped in `#ifdef __cpp_lib_is_constant_evaluated`, like in the other spot at line 144

that's what I did locally but since I only use VS2017, I can't verify that it'll produce the intended outcome for everyone else

vitaut:
You are right, it should be wrapped in `__builtin_is_constant_evaluated`. A PR is welcome!

PR #142, opened 2026-05-25
added a missing feature check ifdef for is_constant_evaluated

archo5:
needed for VS2017+x64 support

fixes issue https://github.com/vitaut/zmij/issues/141

p.s. checked that the `example` project works but none of the test projects build on VS2017 since dragonbox has build errors relating to its various `is_in_range` functions


vitaut:
Merged, thanks

PR #143, opened 2026-06-07
Fix fixed benchmark

TobiSchluter:
Separates out the fixed benchmark added in #137 as requested in review (a long time ago, I had to recover from trying to use AI to find actual improvements and not hallucinated improvements).

Unlike the existing canada benchmark, this benchmark exercises the full exponential range of the fixed format output. The code path for negative exponents is different in that zeros are prefixed and the decimal point isn't inserted inside the digit string. This benchmark ensures that that variant is also exercised.

This uses an equidistribution of exponents and equidistributes the numbers in the range of each exponent. One change relative to the version in #137 is that I also added 50/50 randomness to the sign.

### DIFF
--- a/test/benchmark.cc
+++ b/test/benchmark.cc
@@ -161,6 +161,49 @@ static void run_to_chars_canada(benchmark::State& state,
           benchmark::Counter::kInvert);
 }
 
+// Doubles drawn uniformly from zmij's fixed-notation decimal-exponent range,
+// dec_exp in [-4, 15]: each decade is equally weighted, so the negative-
+// exponent side (which canada.json doesn't cover) gets ~25% of samples. Sized
+// to roughly match canada_numbers for comparable per-iteration cost.
+static const std::vector<double>& get_fixed_range_numbers() {
+  static const std::vector<double> v = [] {
+    constexpr size_t count =
+        sizeof(canada_numbers) / sizeof(canada_numbers[0]);
+    std::mt19937_64 rng(0);
+    std::uniform_int_distribution<int> exp_dist(-4, 15);
+    std::uniform_real_distribution<double> sig_dist(1.0, 10.0);
+    std::vector<double> out;
+    out.reserve(count);
+    for (size_t i = 0; i < count; ++i) {
+      double sig = sig_dist(rng);
+      int e = exp_dist(rng);
+      out.push_back(sig * std::pow(10.0, e));
+    }
+    return out;
+  }();
+  return v;
+}
+
+static void run_to_chars_fixed_range(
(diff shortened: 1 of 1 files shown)


test/benchmark.cc
```diff
@@ -161,6 +161,53 @@ static void run_to_chars_canada(benchmark::State& state,
           benchmark::Counter::kInvert);
 }
 
+// Doubles drawn uniformly from zmij's fixed-notation decimal-exponent range,
+// dec_exp in [-4, 15]: each decade is equally weighted, so the negative-
+// exponent side (which canada.json doesn't cover) gets ~25% of samples. Signs
```
vitaut, line 166: Shouldn't it be ~20%?

vitaut:
Thanks for the PR! Mostly looks good but how different are the results between this benchmark and canada?

test/benchmark.cc
TobiSchluter: well, yes. I guess that is counting -4 to -0 vs 1 to 15, but that would also not make sense. Will correct

TobiSchluter:
Right, I had prepared the numbers but didn't include them when writing the PR. The two benchmarks aren't very different with gcc, but surprisingly with clang fixed-range benchmarks significantly faster, e.g. on my Zen4 with the version in the pull request:

| Build    | zmij canada | zmij fixed_range | dragonbox canada | dragonbox fixed_range |
|----------|------------:|-----------------:|-----------------:|----------------------:|
| g++ v1   |        8.62 |             8.56 |            12.14 |                 18.85 |
| g++ v2   |        7.94 |             7.73 |            12.25 |                 19.12 |
| clang v1 |        8.45 |             7.65 |            11.35 |                 16.87 |
| clang v2 |        7.41 |             6.83 |            11.71 |                 17.15 |

(v1 = SSE2, v2 = SSE 4.2) I mainly included it to prevent myself from accidentally pessimizing the negative exponent case. Dragonbox shows us that this is possible.

PS here are the numbers with the changes from the other #137 patch. I made a few more runs and couldn't repeat the g++v2 fixed_range regression, but I'm not going to tamper with the numbers.
| Build    | zmij canada | zmij fixed_range |
|----------|------------:|-----------------:|
| g++ v1   |        8.60 |             8.66 |
| g++ v2   |        7.76 |             8.23 |
| clang v1 |        8.00 |             7.40 |
| clang v2 |        6.68 |             6.68 |

test/benchmark.cc
vitaut: That's minor, so I can fix it in a follow-up commit. Just wanted to confirm.

TobiSchluter: thanks!

PR #147, opened 2026-08-10
Reduce data cache pressure in dec_digits especially if num_bits is 32 (float type) and optimize algorithms in SSE patch, also it fix many compilation issues when different compilers.

IRainman:
This is an additional for #133


IRainman:
I do not have a macOS or ARM machine. Can you @vitaut please help me? Or I incorrectly understand, and the problem is not inmy code now?

IRainman (commented):
Now the test should work, and the building works fine.

IRainman (commented):
I guess it is. Now I want feedback and recommendations.

vitaut:
Thanks for the PR but I don't think it's ready for review because it combines too many unrelated changes. Please split into small self-contained changes with clear motivation and backed benchmark results for optimizations.

PR #148, opened 2026-08-25
Allow building with older gcc

patrickbkr:
This change fixes zmij to build on older gcc. Without this patch it fails to build on at least gcc 6.3.0 and 5.3.1. This is the error:

    zmij.c:1118:42: error: initializer element is not constant
    #define ZMIJ_SPLAT64(x) {(long long)(x), (long long)(x)}
                                             ^


vitaut:
Thank you

PR #149, opened 2026-08-26
Add minimal CMake install support

miyanyan:
~~Also added cmake options to control examples and tests, default ON~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,12 +1,17 @@
 cmake_minimum_required(VERSION 3.16)
 project(zmij CXX C)
 
+include(GNUInstallDirs)
+
 set(ZMIJ_STANDARD cxx_std_14)
 
 option(ZMIJ_USE_SIMD "Use SIMD instructions" ON)
 
 add_library(zmij zmij.cc zmij.h)
-target_include_directories(zmij PUBLIC .)
+add_library(zmij::zmij ALIAS zmij)
+target_include_directories(
+  zmij PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+              $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 target_compile_features(zmij PRIVATE ${ZMIJ_STANDARD})
 if (NOT ZMIJ_USE_SIMD)
   target_compile_definitions(zmij PRIVATE "ZMIJ_USE_SIMD=0")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -4,16 +4,33 @@ project(zmij CXX C)
...
+if (ZMIJ_BUILD_TESTS)
+  set(CMAKE_CTEST_ARGUMENTS "--output-on-failure")
+  enable_testing()
+  add_subdirectory(test)
+endif ()
 
-set(CMAKE_CTEST_ARGUMENTS "--output-on-failure")
-enable_testing()
-add_subdirectory(test)
+install(TARGETS zmij EXPORT zmij-targets ARCHIVE DESTINATION lib)
```
vitaut: `ARCHIVE DESTINATION` won't work for the shared library and is redundant for static library, let's remove.

CMakeLists.txt
```diff
@@ -4,16 +4,33 @@ project(zmij CXX C)
...
+install(FILES zmij.h zmij-to-chars.h DESTINATION include)
```
vitaut: `include` -> `${CMAKE_INSTALL_INCLUDEDIR}`

CMakeLists.txt
```diff
@@ -4,16 +4,33 @@ project(zmij CXX C)
...
+install(
+  EXPORT zmij-targets
+  FILE zmij-config.cmake
+  NAMESPACE zmij::
+  DESTINATION share/zmij)
```
vitaut: `share/zmij` -> `${CMAKE_INSTALL_LIBDIR}/cmake/zmij`

CMakeLists.txt
```diff
@@ -4,16 +4,33 @@ project(zmij CXX C)
 set(ZMIJ_STANDARD cxx_std_14)
 
 option(ZMIJ_USE_SIMD "Use SIMD instructions" ON)
+option(ZMIJ_BUILD_EXAMPLE "Build the example" ON)
+option(ZMIJ_BUILD_TESTS "Build the tests" ON)
```
vitaut: This seems unrelated to installation, let's remove.

vitaut (changes requested):
Thanks for the PR.

CMakeLists.txt
miyanyan: I will open another pr to add these options, then I can add zmij to vcpkg

vitaut (approved):

vitaut:
Merged, thank you!

PR #151, opened 2026-08-29
optimize SSE path by using more optimized instruction to registers un…

IRainman:
…load and remove memmove from hot patch.

### DIFF
--- a/zmij.cc
+++ b/zmij.cc
@@ -1159,8 +1159,8 @@ ZMIJ_INLINE auto to_digits<32>(uint64_t value,
 ZMIJ_INLINE void write_digits(char* buffer, dec_digits<64>::digits_type digits,
                               bool drop_leading_zero, const data& d) noexcept {
   if (!ZMIJ_USE_NEON && !ZMIJ_USE_SSE4_1) {
+    digits = digits >> (drop_leading_zero * sizeof(digits));
     memcpy(buffer, &digits, sizeof(digits));
-    memmove(buffer, buffer + drop_leading_zero, sizeof(digits));
     return;
   }
 #if ZMIJ_USE_NEON
@@ -1172,13 +1172,16 @@ ZMIJ_INLINE void write_digits(char* buffer, dec_digits<64>::digits_type digits,
       reinterpret_cast<const __m128i*>(d.shift_shuffle + drop_leading_zero));
   _mm_storeu_si128(reinterpret_cast<__m128i*>(buffer),
                    _mm_shuffle_epi8(digits, shuffle));
+#elif ZMIJ_USE_SSE
+  _mm_storeu_si128(reinterpret_cast<__m128i*>(buffer + drop_leading_zero),
+                   digits);
 #endif
 }
 
 ZMIJ_INLINE void write_digits(char* buffer, uint64_t digits,
                               bool drop_leading_zero, const data&) noexcept {
+  digits = digits >> (drop_leading_zero * sizeof(digits));
   memcpy(buffer, &digits, sizeof(digits));
-  memmove(buffer, buffer + drop_leading_zero, sizeof(digits));
 }
 
 ZMIJ_INLINE auto write_scientific_simd(char* buffer, const dec_digits<32>& dig,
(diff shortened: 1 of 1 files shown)


IRainman (commented):
it is

zmij.cc
```diff
@@ -1172,13 +1172,16 @@ ZMIJ_INLINE void write_digits(char* buffer, dec_digits<64>::digits_type digits,
...
                    _mm_shuffle_epi8(digits, shuffle));
+#elif ZMIJ_USE_SSE
+  _mm_storeu_si128(reinterpret_cast<__m128i*>(buffer + drop_leading_zero),
+                   digits);
 #endif
 }
 
 ZMIJ_INLINE void write_digits(char* buffer, uint64_t digits,
                               bool drop_leading_zero, const data&) noexcept {
+  digits = digits >> (drop_leading_zero * sizeof(digits));
```
ruema, line 1183: actually, you want to shift by 8 bits.
`digits >>= drop_leading_zero * 8;`

zmij.cc
```diff
@@ -1324,7 +1327,11 @@ ZMIJ_INLINE auto write_scientific_digits(char* buffer,
                                          dec_digits<64>::digits_type digits,
                                          unsigned lo, int num_digits,
                                          int dec_exp) noexcept -> char* {
+#if ZMIJ_USE_SSE || ZMIJ_USE_SSE4_1
+  _mm_storeu_si128(reinterpret_cast<__m128i*>(buffer + 1), digits);
```
ruema, line 1331: The 16 byte memcpy call is replaced by the appropriate sse-move operation by the compiler.

zmij.cc
```diff
@@ -1172,13 +1172,16 @@ ZMIJ_INLINE void write_digits(char* buffer, dec_digits<64>::digits_type digits,
       reinterpret_cast<const __m128i*>(d.shift_shuffle + drop_leading_zero));
   _mm_storeu_si128(reinterpret_cast<__m128i*>(buffer),
                    _mm_shuffle_epi8(digits, shuffle));
+#elif ZMIJ_USE_SSE
+  _mm_storeu_si128(reinterpret_cast<__m128i*>(buffer + drop_leading_zero),
+                   digits);
```
vitaut, line 1177: This branch is currently unreachable for SSE2 because the if (!ZMIJ_USE_NEON && !ZMIJ_USE_SSE4_1) above returns first. Also, if made reachable, storing at buffer + drop_leading_zero doesn't seem equivalent to memmove(buffer, buffer + drop_leading_zero, ...): it shifts the output in the opposite direction.

zmij.cc
vitaut: Indeed: https://godbolt.org/z/xYE6GqWxf

vitaut (changes requested):

zmij.cc
```diff
@@ -1172,13 +1172,16 @@ ZMIJ_INLINE void write_digits(char* buffer, dec_digits<64>::digits_type digits,
...
 #endif
 }
 
 ZMIJ_INLINE void write_digits(char* buffer, uint64_t digits,
                               bool drop_leading_zero, const data&) noexcept {
+  digits = digits >> (drop_leading_zero * sizeof(digits));
   memcpy(buffer, &digits, sizeof(digits));
-  memmove(buffer, buffer + drop_leading_zero, sizeof(digits));
 }
```
vitaut, line 1185: This change is reasonable but it should be 8, not sizeof(digits) and you also need to handle big endian, something like 

```
unsigned shift = unsigned(drop_leading_zero) * 8;
digits = is_big_endian ? digits << shift : digits >> shift;
memcpy(buffer, &digits, sizeof(digits));
```

IRainman: MSVC only optimizes memcpu in O2, not in O1 https://godbolt.org/z/Mbnr33K8M

vitaut: Then whoever cares about perf should compile with O2. I don't think we need to do any heroics for specific MSVC configurations.

AlexGuteniev: > MSVC only optimizes memcpu in O2, not in O1 https://godbolt.org/z/Mbnr33K8M

To be more precise it is `/Oi` that is implied with `/O2`, and it can be also be overridden with `#pragma intrinsic(memcpy)` to optimize in any mode.

Still you should just compile with `/O2` if you care about perf.

TobiSchluter:
It would be nice to see benchmark results for claimed performance improvements.

vitaut:
Merged a fixed version of the write_digits optimization in https://github.com/vitaut/zmij/commit/a24bf7fce0f4fdee60c0153529b74bb537552b7b, thanks.

zmij.cc
IRainman: > just compile with `/O2` if you care about perf.

I always compile code with maximum optimizations, but it's not the case here.

PR #152, opened 2026-08-29
Reduce dec_digits<32> size

IRainman:


IRainman (commented):
it is

vitaut:
I don't think this matters because this is a short-lived object and its parts end up being passed in registers but I guess it won't hurt getting rid of the padding either. Thanks.

PR #153, opened 2026-08-30
Make tests and example optional

miyanyan:
Follow-up to #149.


vitaut:
Thanks for the PR but could you provide some details about the motivation?

miyanyan:
> Thanks for the PR but could you provide some details about the motivation?

Thanks for the review. The main motivation is to make zmij available through vcpkg so that users can install and integrate it more easily. When zmij is consumed as a library, users should not be required to build its tests and examples. These options allow package managers and downstream projects to disable them, much like the `FMT_TEST` option in fmt

CMakeLists.txt
```diff
@@ -6,6 +6,8 @@ include(GNUInstallDirs)
 set(ZMIJ_STANDARD cxx_std_14)
 
 option(ZMIJ_USE_SIMD "Use SIMD instructions" ON)
+option(ZMIJ_BUILD_TESTS "Build tests" ON)
+option(ZMIJ_BUILD_EXAMPLES "Build examples" ON)
```
vitaut: Let's rename to `ZMIJ_TEST` and `ZMIJ_EXAMPLE` to follow fmt's convention and since this is about generating targets, not building. And update the descriptions accordingly, e.g. "Generate the test targets"

vitaut (changes requested):

CMakeLists.txt
miyanyan: Done

vitaut (approved):

vitaut:
Merged, thank you

PR #156, opened 2026-09-10
Itoa implementations for NEON, SSE2, SSE4.1, AVX2

TobiSchluter:
Hi Victor,

I've been watching your progress, and it's been impressive. What I did, isn't so impressive and it took me a while to get around to this, but I'm finally proposing a pull request for the `itoa` stuff.  Headlines first: branch-free (except 128bits) implementations of integer itoa (signed and unsigned <= 32bit, 64bit, and 128bit where available). Predictable runtime down to under 1ns on M5 neon. Winner on almost all benchmarks I could think of.

In order to avoid crashing with your changes, and making the review a bit easier by not interleaving the two, I've split the integer handling into a separate file. Mainly because I reuse `to_bcd8` which counts the trailing zero some unnecessary stuff is present, but I think the unrelated stuff is easier to ignore this way.

What is this optimized for? The idea I had was that the tpyical, performance-aware user will convert one number, and then either copy the string somewhere else or continue inserting characters after each converted integer. Notably, they won't actually look at the emitted bytes directly, and only care about the length of the emitted string which the CPU's out-of-order implementation will fill in while the program continues. So the first thing to optimize is the length evaluation.  Here I used the algorithm found in fmt with three modifications:

1. the algorithm used by fmt for 32bit integers leaves a lot bits unused and can easily be extended in range. For purposes of u128 conversion we need the range up to 1e16, and thus this is extended as such
2. the algorithm for u64 uses two table lookups. In fmt's variant they happen sequentially: the result of the first lookup is used for the second lookup. But at the cost of a larger table one can both lookups simultaneously. Depending on whether we are otpimizing for size or not we use one or the other.
3. the required bitcounting wants `clz` but on old intel CPUs this is implemented as `btz` followed by `xor 63`. REversing the table order makes the `xor` unnecessary. Depending on the compile target the optimal sequence is chosen. The tables themselves are evaluated at compile time.

As for the conversions themselves, it is perhaps worth pointing out that a decimal number string always takes more than twice the space than the binary number. An 32bit number takes up to ten bytes, a 64bit number up to 20, a 128bit number up to 39. Obviously, these numbers don't map well to the registers which are 8, 16 or (AVX2) 32 bytes wide. Therefore, the numbers need to be split, and because SIMD doesn't have full-width 64bit multiplies, they actually need to be split several times. Most of the work went into identifying the optimal splitting sequences. Add to that that on Intel there are several generations of SIMD instructions, that perform differently across different machines, and you will understand that it took some time to come up with what I think are the best sequences.

<details>
<summary>
AI summary of the digit-group splits in the zmij integer conversion
</summary>

Every SIMD tier shares the same 16-digit kernel underneath: the GPR side does
`v / 1e8` to get two 8-digit halves into two 64-bit lanes, then the vector side
splits each lane 8 → 4+4 (`mul_epu32` by the 1e4 reciprocal), 4 → 2+2
(`mulhi_epu16` by the 100 reciprocal), 2 → 1+1 (`mulhi` by the 10 reciprocal),
and a final shuffle reverses and trims leading zeros. The differences below are
about what is fed into that kernel and what is handled outside it.

## u32 (up to 10 digits)

| Tier | Split | Where |
|---|---|---|
| scalar and SSE2 | 8 + 2 | GPR: `v/100` and `v/1e6` in parallel, feeding `to_bcd8_split` (SWAR 3-step on scalar, SSE2 `to_bcd_4x4` on SSE2), right-shifted to trim and stored as 8 bytes; low 2 digits via the `digits2` table |
| SSE4.1 and AVX2 | 4 + 4 + 2 in one vector | GPR: `v/100` and `v/1e6`; the 8-digit part is packed to two 4-digit lanes with one madd, the 2-digit remainder goes in a third lane. One `to_ascii_4x4` pass and one pshufb from `revalign_shuffle10`, one 16-byte store |
| NEON | 16 (full kernel) | No narrow kernel. `v/1e8` (0..42) packed with the remainder by one madd, full 16-digit body, one store. Signed i32 folds the `-` into the same store via a widened shuffle window and a `sign_bias` row |

## u64 (up to 20 digits)

| Tier | Split | Where |
|---|---|---|
| scalar | 4 + 8 + 8 | `v/1e8` and `v/1e16` (parallel). Top ≤ 4 digits via divmod100 + two `digits2` lookups, mid and low via `to_bcd8` SWAR. Assembled right-aligned in a 48-byte buffer, one 20-byte copy at 24 − len |
| SSE2 | 16 + 4 | `v/1e4` in GPR; body = high if v ≥ 1e16 else v, so body < 1e16. Body pre-scaled by 10^(16−len) so the kernel output is already left-aligned, no shuffle. Low 4 digits always written after via two `digits2` lookups, counted only if v ≥ 1e16 |
| SSE4.1 generic, AVX2 generic | 16 + 4 | Same peel as SSE2. Body goes through the pshufb `revalign_shuffle` body instead of the scale trick; low 4 digits via `digits2` |
| SSE4.1 + `ZMIJ_USE_U64_SPLIT12` (Zen 5 default) | 12 + 8 | `v/1e16` and `v/1e8` as two independent reciprocal multiplies, no chaining. Lanes (q16, q8 mod 1e8) go through the 16-digit kernel with an unclamped shuffle offset of 24 − c, which emits all-padding for v < 1e8. Low 8 digits via `split10k` + `to_ascii_4x4` + a `revalign_shuffle` window, 8-byte store at out + klen |
| AVX2 + `ZMIJ_USE_AVX2_U64_FP` (Zen 5 default) | 4 + 16 | Same two independent divides. Head q16 (≤ 4 digits) via the float reciprocal kernel `to_ascii4_ps` on the FP ports, trimmed by the `peel4_pack` window, 4-byte store. Remainder lanes through `itoa_body_lanes` at out + hlen, whose 16-byte store overwrites the head's garbage |
| NEON | 16 + 4 | Same peel as SSE2, `v/1e4` via umulh with a `static_data` reciprocal, compare against 1e16 − 1 from the same ldp. Low 4 via a /100 reciprocal and two `digits2` lookups |

## u128 (only above `UINT64_MAX`, else the u64 path)

First peel: a 128-bit reciprocal divmod by 1e16 gives the low 16 digits and a
quotient. If the quotient is below 1e16 (19 to 32 digits total) the split is
top ≤ 16 trimmed + low 16 padded. Otherwise a second narrow divmod (a 64-bit
reciprocal for 5^16 on quotient >> 16) gives top ≤ 7 + mid 16 + low 16.

| Tier | 19–32 digits | 33–39 digits |
|---|---|---|
| scalar | 8+8 \| 8+8, four `to_bcd8`, one 32-byte copy from a 64-byte buffer | 8 \| 8+8 \| 8+8, five `to_bcd8`, one 40-byte copy |
| SSE2 | 16 trimmed (scaled body) + 16 padded (`to_ascii16` via pshufd, no bswap) | `itoa_head7`: one 8-digit BCD group, shifted and stored as 8 bytes, then 16 + 16 padded |
| SSE4.1 | 16 trimmed (pshufb body) + 16 padded (bswap shuffle) | top through the full 16-digit body, then 16 + 16 padded, three passes |
| AVX2 | 16 + 16 in one 256-bit pass, lane 0 gets the trim shuffle, lane 1 the plain reversal, the two lanes stored at out and out + hlen | `itoa_top8`: four base-100 blocks via float reciprocals, one SWAR /10 split, one pshufb trim, 8-byte store; then 32 padded in one 256-bit pass |
| NEON | 16 trimmed + 16 padded (offset 0 into `revalign_shuffle` is the pure reversal) | top through the 16-digit body, then 16 + 16 padded in a noinline `itoa_body32_pad` |

## Signed wrappers

i128 first checks whether the value fits in i64 and takes that path. For
everything else the sign is written first and the unsigned path runs, except
i32 on NEON, which has the dedicated single-store body above.

## Note on AVX2

AVX2 without the Zen 5 tuning flag does not get its own u64 split. The first
branch in `itoa` requires either `ZMIJ_USE_AVX2_U64_FP` or
`ZMIJ_USE_U64_SPLIT12`, so a plain `-mavx2` build falls to the same 16 + 4 peel
as generic SSE4.1, and only the u128 paths differ.
</details>

As for the floating point conversions, AVX2 turned out to be quite useless. I managed to find two uses though: for large u128, the low 32bit digits can be handled in one AVX2 chain. Additionally, on architectures with lots of wide multipliers (Zen5) it is possible to use the floating point fma instruction to convert 4 digits in a minimum of operations. This is used for the 4 remaining digits after a 20 digits 64-bit number is split in to 16+4 digits.

As for benchmarks, in my [fork of the itoa-benchmark repo](https://github.com/TobiSchluter/itoa-benchmark) you can find a number of low-level benchmarks. The current plots are for the same code as the one in this PR. For the PR, I chose to define a smaller set of benchmarks, and I looked for some more inspiration. Champagne and Lemire used a "Twitter JSON" dataset to benchmark their AVX512 itoa code. Unfortunately, that sample only contains some 2000 numbers -- an amount the branch prediction in a modern CPU easily trivializes, so I added a Markov-chain generator that is trained on that sample but generates much longer samples. This is the twitter benchmark, each number is converted, `", "` is inserted after the number and the next number is converted. For this benchmark, I added a variety that always uses u64 conversion and one that picks the conversion based on the size of the number (most are smaller than <2e32, and so it is fairly predictable). The labels are zmij and zmij32 , respectively.

Another benchmark converts i8 and u8 numbers, either as 32bit or as 64bit integers. This is a bit of a worst-case, as it can be handled trivially via a table, but still the 32bit code wins against fmt (the baseline for comparison). This also the only loss against fmt in the benchmarks.

Finally, and maybe not really applicable to the real world, I measured log-uniform distributions. Here we routinely see speed increases which reach more than an order of magnitude over fmt.

<details>
<summary>Detailed benchmark results</summary>

Note that the table show a few variants not included in the PR for the twitter benchmark: "iid" is a flat generator which picks from the set of twitter.json numbers. mk1 is a Markov-chain generator, i.e. the next number is picked on the conditional distribution given by the length of the current number. The mk2 is the default included in the repo, and it makes the number's length depend on the two preceding numbers.

All values are average nanoseconds per conversion. Our columns also contain the relative speed up compared to fmt.

## AMD Ryzen 9 9950X (Zen 5), gcc 16

| benchmark | fmt | scalar | v1 | v2 | v3 | native |
|---|---|---|---|---|---|---|
| json_twitter_iid_u64 | 6.90 | 5.12 (+35%) | 3.92 (+76%) | 2.71 (+155%) | 2.60 (+165%) | 2.65 (+160%) |
| json_twitter_iid_u64 (zmij32) | 6.90 | 3.86 (+79%) | 3.61 (+91%) | 2.83 (+144%) | 2.82 (+145%) | 2.72 (+154%) |
| json_twitter_mk1_u64 | 5.54 | 5.09 (+9%) | 3.91 (+42%) | 2.71 (+104%) | 2.59 (+114%) | 2.69 (+106%) |
| json_twitter_mk1_u64 (zmij32) | 5.54 | 3.87 (+43%) | 3.51 (+58%) | 2.80 (+98%) | 2.70 (+105%) | 2.69 (+106%) |
| json_twitter_mk2_u64 | 3.86 | 5.09 (-24%) | 3.91 (-1%) | 2.74 (+41%) | 2.57 (+50%) | 2.68 (+44%) |
| json_twitter_mk2_u64 (zmij32) | 3.86 | 3.85 (+0%) | 3.61 (+7%) | 2.83 (+37%) | 2.71 (+43%) | 2.72 (+42%) |
| json_u8_as_u32 | 3.25 | 2.55 (+28%) | 2.25 (+45%) | 1.65 (+97%) | 1.52 (+114%) | 1.54 (+111%) |
| json_i8_as_i32 | 4.52 | 2.79 (+62%) | 2.47 (+83%) | 1.95 (+132%) | 1.82 (+149%) | 1.78 (+155%) |
| json_u8_as_u64 | 3.22 | 5.09 (-37%) | 3.90 (-17%) | 2.72 (+19%) | 2.55 (+27%) | 2.63 (+23%) |
| json_i8_as_i64 | 4.77 | 5.38 (-11%) | 4.18 (+14%) | 2.97 (+61%) | 2.93 (+63%) | 2.80 (+70%) |
| log_uniform_u32 | 8.61 | 2.17 (+297%) | 1.85 (+366%) | 1.35 (+536%) | 1.33 (+549%) | 1.16 (+640%) |
| log_uniform_u64 | 11.13 | 4.96 (+124%) | 3.35 (+232%) | 2.31 (+382%) | 2.16 (+415%) | 2.26 (+391%) |
| log_uniform_i32 | 8.29 | 2.53 (+228%) | 2.12 (+291%) | 1.50 (+453%) | 1.44 (+476%) | 1.40 (+493%) |
| log_uniform_i64 | 10.96 | 5.16 (+113%) | 3.58 (+206%) | 2.68 (+309%) | 2.41 (+356%) | 2.42 (+353%) |
| log_uniform_u128 | 101.95 | 10.67 (+856%) | 6.80 (+1400%) | 5.95 (+1615%) | 5.60 (+1722%) | 5.96 (+1612%) |
| log_uniform_i128 | 68.24 | 9.89 (+590%) | 7.91 (+763%) | 7.27 (+839%) | 7.05 (+868%) | 8.09 (+743%) |

## AMD Ryzen 9 9950X (Zen 5), clang 21

| benchmark | fmt | scalar | v1 | v2 | v3 | native |
|---|---|---|---|---|---|---|
| json_twitter_iid_u64 | 7.51 | 5.20 (+44%) | 3.98 (+89%) | 2.79 (+170%) | 2.59 (+190%) | 2.56 (+194%) |
| json_twitter_iid_u64 (zmij32) | 7.51 | 3.86 (+94%) | 3.57 (+110%) | 2.94 (+156%) | 2.80 (+169%) | 2.76 (+172%) |
| json_twitter_mk1_u64 | 6.00 | 5.25 (+14%) | 3.99 (+50%) | 2.81 (+114%) | 2.58 (+133%) | 2.55 (+135%) |
| json_twitter_mk1_u64 (zmij32) | 6.00 | 3.89 (+54%) | 3.55 (+69%) | 2.95 (+103%) | 2.83 (+112%) | 2.76 (+117%) |
| json_twitter_mk2_u64 | 4.13 | 5.28 (-22%) | 4.00 (+3%) | 2.90 (+42%) | 2.57 (+61%) | 2.55 (+62%) |
| json_twitter_mk2_u64 (zmij32) | 4.13 | 3.90 (+6%) | 3.60 (+15%) | 2.95 (+40%) | 2.83 (+46%) | 2.89 (+43%) |
| json_u8_as_u32 | 3.85 | 2.57 (+50%) | 2.21 (+74%) | 1.73 (+123%) | 1.60 (+141%) | 1.61 (+140%) |
| json_i8_as_i32 | 2.96 | 2.83 (+5%) | 2.52 (+18%) | 1.97 (+50%) | 1.86 (+59%) | 1.82 (+62%) |
| json_u8_as_u64 | 3.60 | 5.24 (-31%) | 3.90 (-8%) | 2.82 (+28%) | 2.57 (+40%) | 2.55 (+41%) |
| json_i8_as_i64 | 3.00 | 5.42 (-45%) | 4.30 (-30%) | 3.16 (-5%) | 2.84 (+5%) | 2.86 (+5%) |
| log_uniform_u32 | 8.66 | 2.11 (+310%) | 1.76 (+391%) | 1.27 (+582%) | 1.13 (+667%) | 1.16 (+644%) |
| log_uniform_u64 | 11.23 | 4.84 (+132%) | 3.34 (+237%) | 2.30 (+389%) | 2.02 (+455%) | 2.01 (+458%) |
| log_uniform_i32 | 8.14 | 2.44 (+234%) | 2.00 (+307%) | 1.48 (+448%) | 1.37 (+494%) | 1.35 (+501%) |
| log_uniform_i64 | 11.10 | 5.09 (+118%) | 3.53 (+215%) | 2.54 (+338%) | 2.29 (+385%) | 2.29 (+386%) |
| log_uniform_u128 | 78.98 | 11.95 (+561%) | 6.16 (+1182%) | 5.46 (+1348%) | 5.00 (+1479%) | 4.92 (+1505%) |
| log_uniform_i128 | 47.87 | 10.28 (+366%) | 7.38 (+549%) | 7.59 (+530%) | 6.65 (+620%) | 6.47 (+640%) |

## Apple M5, Apple clang 21

| benchmark | fmt | scalar | neon |
|---|---|---|---|
| json_twitter_iid_u64 | 5.67 | 3.23 (+76%) | 2.26 (+151%) |
| json_twitter_iid_u64 (zmij32) | 5.67 | 2.95 (+92%) | 2.26 (+151%) |
| json_twitter_mk1_u64 | 4.50 | 3.26 (+38%) | 2.26 (+99%) |
| json_twitter_mk1_u64 (zmij32) | 4.50 | 2.98 (+51%) | 2.30 (+95%) |
| json_twitter_mk2_u64 | 3.17 | 3.28 (-3%) | 2.26 (+40%) |
| json_twitter_mk2_u64 (zmij32) | 3.17 | 3.01 (+5%) | 2.26 (+40%) |
| json_u8_as_u32 | 2.66 | 1.99 (+34%) | 1.94 (+38%) |
| json_i8_as_i32 | 2.47 | 2.21 (+12%) | 1.95 (+27%) |
| json_u8_as_u64 | 2.59 | 3.29 (-21%) | 2.26 (+14%) |
| json_i8_as_i64 | 2.34 | 3.48 (-33%) | 2.49 (-6%) |
| log_uniform_u32 | 6.70 | 1.66 (+303%) | 0.97 (+591%) |
| log_uniform_u64 | 9.61 | 3.06 (+214%) | 1.51 (+536%) |
| log_uniform_i32 | 6.28 | 1.83 (+244%) | 1.16 (+444%) |
| log_uniform_i64 | 9.70 | 3.26 (+198%) | 1.62 (+498%) |
| log_uniform_u128 | 136.13 | 9.39 (+1350%) | 4.26 (+3098%) |
| log_uniform_i128 | 85.27 | 9.23 (+824%) | 5.73 (+1387%) |

## Apple M1 (192.168.1.98), Apple clang 21

| benchmark | fmt | scalar | neon |
|---|---|---|---|
| json_twitter_iid_u64 | 8.77 | 4.95 (+77%) | 4.26 (+106%) |
| json_twitter_iid_u64 (zmij32) | 8.77 | 4.89 (+79%) | 4.00 (+119%) |
| json_twitter_mk1_u64 | 6.95 | 4.99 (+39%) | 4.40 (+58%) |
| json_twitter_mk1_u64 (zmij32) | 6.95 | 4.95 (+41%) | 4.05 (+72%) |
| json_twitter_mk2_u64 | 5.04 | 4.95 (+2%) | 4.39 (+15%) |
| json_twitter_mk2_u64 (zmij32) | 5.04 | 4.91 (+3%) | 4.03 (+25%) |
| json_u8_as_u32 | 4.45 | 4.01 (+11%) | 3.69 (+21%) |
| json_i8_as_i32 | 4.21 | 4.72 (-11%) | 3.70 (+14%) |
| json_u8_as_u64 | 4.26 | 4.94 (-14%) | 4.39 (-3%) |
| json_i8_as_i64 | 4.43 | 5.45 (-19%) | 4.88 (-9%) |
| log_uniform_u32 | 10.86 | 2.81 (+286%) | 1.61 (+573%) |
| log_uniform_u64 | 13.88 | 4.67 (+197%) | 2.50 (+454%) |
| log_uniform_i32 | 10.20 | 3.09 (+230%) | 1.87 (+444%) |
| log_uniform_i64 | 14.12 | 4.88 (+189%) | 2.67 (+429%) |
| log_uniform_u128 | 223.66 | 14.19 (+1477%) | 6.67 (+3255%) |
| log_uniform_i128 | 138.11 | 13.65 (+912%) | 8.48 (+1528%) |

## Intel Core i7-1165G7 (Tiger Lake), gcc 16

| benchmark | fmt | scalar | v1 | v2 | v3 | native |
|---|---|---|---|---|---|---|
| json_twitter_iid_u64 (zmij32) | n/a | 8.08 | 8.41 | 7.49 | 6.78 | 6.90 |
| json_twitter_mk1_u64 (zmij32) | n/a | 9.94 | 8.36 | 7.03 | 6.63 | 7.11 |
| json_twitter_mk2_u64 | 8.87 | 10.52 (-16%) | 9.61 (-8%) | 7.76 (+14%) | 7.48 (+19%) | 7.28 (+22%) |
| json_twitter_mk2_u64 (zmij32) | 8.87 | 8.09 (+10%) | 9.10 (-3%) | 7.25 (+22%) | 6.86 (+29%) | 7.76 (+14%) |
| json_u8_as_u32 | 6.26 | 5.46 (+15%) | 6.40 (-2%) | 5.07 (+23%) | 4.95 (+26%) | 5.01 (+25%) |
| json_i8_as_i32 | 8.60 | 6.50 (+32%) | 6.50 (+32%) | 6.32 (+36%) | 5.56 (+55%) | 5.94 (+45%) |
| json_u8_as_u64 | 6.26 | 11.34 (-45%) | 8.61 (-27%) | 8.31 (-25%) | 8.18 (-23%) | 7.28 (-14%) |
| json_i8_as_i64 | 8.42 | 11.14 (-24%) | 9.26 (-9%) | 8.55 (-2%) | 7.88 (+7%) | 7.76 (+8%) |
| log_uniform_u32 | 16.94 | 5.21 (+225%) | 5.32 (+219%) | 4.85 (+250%) | 4.01 (+323%) | 3.83 (+343%) |
| log_uniform_u64 | 22.47 | 9.77 (+130%) | 8.13 (+176%) | 7.42 (+203%) | 8.09 (+178%) | 6.61 (+240%) |
| log_uniform_i32 | 14.73 | 5.51 (+167%) | 5.79 (+154%) | 5.12 (+188%) | 4.94 (+198%) | 5.29 (+179%) |
| log_uniform_i64 | 23.01 | 11.01 (+109%) | 9.71 (+137%) | 7.68 (+199%) | 7.28 (+216%) | 7.27 (+217%) |
| log_uniform_u128 | 243.47 | 21.23 (+1047%) | 20.12 (+1110%) | 18.99 (+1182%) | 14.21 (+1613%) | 16.06 (+1416%) |
| log_uniform_i128 | 136.03 | 21.83 (+523%) | 18.11 (+651%) | 18.50 (+635%) | 18.00 (+656%) | 18.54 (+634%) |

## Intel Core i7-1165G7 (Tiger Lake), clang 21

| benchmark | fmt | scalar | v1 | v2 | v3 | native |
|---|---|---|---|---|---|---|
| json_twitter_iid_u64 (zmij32) | n/a | 8.80 | 8.86 | 7.70 | 8.18 | 7.62 |
| json_twitter_mk1_u64 (zmij32) | n/a | 8.61 | 8.94 | 7.75 | 7.32 | 7.28 |
| json_twitter_mk2_u64 | 9.30 | 11.39 (-18%) | 9.62 (-3%) | 8.72 (+7%) | 8.41 (+11%) | 10.82 (-14%) |
| json_twitter_mk2_u64 (zmij32) | 9.30 | 8.48 (+10%) | 8.56 (+9%) | 7.42 (+25%) | 7.34 (+27%) | 7.64 (+22%) |
| json_u8_as_u32 | 7.76 | 5.81 (+33%) | 6.11 (+27%) | 5.07 (+53%) | 4.75 (+63%) | 5.08 (+53%) |
| json_i8_as_i32 | 6.29 | 6.79 (-7%) | 6.99 (-10%) | 6.41 (-2%) | 5.78 (+9%) | 5.55 (+13%) |
| json_u8_as_u64 | 7.64 | 10.20 (-25%) | 9.67 (-21%) | 8.15 (-6%) | 9.16 (-17%) | 7.82 (-2%) |
| json_i8_as_i64 | 6.42 | 12.41 (-48%) | 10.16 (-37%) | 10.31 (-38%) | 9.48 (-32%) | 9.61 (-33%) |
| log_uniform_u32 | 14.77 | 5.19 (+184%) | 5.34 (+177%) | 4.22 (+250%) | 3.96 (+273%) | 4.13 (+258%) |
| log_uniform_u64 | 22.49 | 10.37 (+117%) | 8.43 (+167%) | 8.53 (+164%) | 8.04 (+180%) | 7.19 (+213%) |
| log_uniform_i32 | 14.55 | 5.48 (+165%) | 5.64 (+158%) | 4.64 (+213%) | 4.67 (+212%) | 4.63 (+214%) |
| log_uniform_i64 | 22.01 | 9.81 (+124%) | 8.86 (+148%) | 8.53 (+158%) | 7.93 (+178%) | 8.84 (+149%) |
| log_uniform_u128 | 154.24 | 22.66 (+581%) | 17.94 (+760%) | 15.85 (+873%) | 15.71 (+882%) | 15.91 (+869%) |
| log_uniform_i128 | 107.93 | 21.70 (+397%) | 23.12 (+367%) | 18.22 (+493%) | 16.23 (+565%) | 15.67 (+589%) |

## 13th Gen Intel Core i9-13900KF (Raptor Lake), gcc 16

benchmark | fmt | scalar | v1 | v2 | v3 | native
-- | -- | -- | -- | -- | -- | --
json_twitter_iid_u64 | 7.14 | 6.07 (+18%) | 3.90 (+83%) | 3.69 (+93%) | 3.63 (+97%) | 3.66 (+95%)
json_twitter_iid_u64 (zmij32) | 7.14 | 3.76 (+90%) | 3.68 (+94%) | 3.08 (+132%) | 2.99 (+139%) | 3.00 (+138%)
json_twitter_mk1_u64 | 5.66 | 6.04 (-6%) | 3.90 (+45%) | 3.72 (+52%) | 3.66 (+55%) | 3.63 (+56%)
json_twitter_mk1_u64 (zmij32) | 5.66 | 3.79 (+49%) | 3.70 (+53%) | 3.09 (+83%) | 2.99 (+89%) | 3.00 (+88%)
json_twitter_mk2_u64 | 3.97 | 6.08 (-35%) | 3.90 (+2%) | 3.70 (+7%) | 3.65 (+9%) | 3.65 (+9%)
json_twitter_mk2_u64 (zmij32) | 3.97 | 3.84 (+3%) | 3.71 (+7%) | 3.08 (+29%) | 2.99 (+32%) | 3.02 (+31%)
json_u8_as_u32 | 3.26 | 2.56 (+28%) | 2.47 (+32%) | 1.92 (+70%) | 1.85 (+77%) | 1.85 (+76%)
json_i8_as_i32 | 4.55 | 2.80 (+62%) | 2.72 (+68%) | 2.14 (+112%) | 2.08 (+119%) | 2.07 (+120%)
json_u8_as_u64 | 3.39 | 6.01 (-44%) | 3.88 (-13%) | 3.70 (-8%) | 3.67 (-8%) | 3.63 (-7%)
json_i8_as_i64 | 5.05 | 6.40 (-21%) | 4.25 (+19%) | 3.95 (+28%) | 3.90 (+30%) | 3.93 (+29%)
log_uniform_u32 | 12.11 | 2.56 (+374%) | 2.19 (+452%) | 1.71 (+606%) | 1.64 (+639%) | 1.68 (+622%)
log_uniform_u64 | 16.43 | 5.58 (+194%) | 3.51 (+368%) | 3.40 (+382%) | 3.31 (+397%) | 3.30 (+398%)
log_uniform_i32 | 14.00 | 2.61 (+437%) | 2.43 (+477%) | 1.88 (+646%) | 1.84 (+663%) | 1.81 (+672%)
log_uniform_i64 | 19.08 | 5.85 (+226%) | 3.83 (+398%) | 3.64 (+424%) | 3.54 (+439%) | 3.57 (+435%)
log_uniform_u128 | 107.37 | 10.36 (+937%) | 7.69 (+1297%) | 7.11 (+1409%) | 6.96 (+1443%) | 6.96 (+1442%)
log_uniform_i128 | 71.70 | 10.10 (+610%) | 8.55 (+739%) | 8.42 (+751%) | 8.27 (+767%) | 8.26 (+768%)

## 13th Gen Intel Core i9-13900KF (Raptor Lake), clang 21

benchmark | fmt | scalar | v1 | v2 | v3 | native
-- | -- | -- | -- | -- | -- | --
json_twitter_iid_u64 | 8.13 | 6.15 (+32%) | 3.98 (+104%) | 3.64 (+124%) | 3.61 (+125%) | 3.62 (+125%)
json_twitter_iid_u64 (zmij32) | 8.13 | 3.82 (+113%) | 3.67 (+122%) | 3.22 (+152%) | 3.07 (+165%) | 3.13 (+160%)
json_twitter_mk1_u64 | 6.56 | 6.16 (+6%) | 3.96 (+66%) | 3.68 (+78%) | 3.61 (+82%) | 3.64 (+80%)
json_twitter_mk1_u64 (zmij32) | 6.56 | 3.79 (+73%) | 3.68 (+78%) | 3.18 (+106%) | 3.09 (+112%) | 3.17 (+107%)
json_twitter_mk2_u64 | 4.51 | 6.15 (-27%) | 3.99 (+13%) | 3.64 (+24%) | 3.63 (+24%) | 3.65 (+24%)
json_twitter_mk2_u64 (zmij32) | 4.51 | 3.82 (+18%) | 3.67 (+23%) | 3.23 (+40%) | 3.08 (+46%) | 3.12 (+44%)
json_u8_as_u32 | 3.80 | 2.56 (+48%) | 2.42 (+57%) | 1.93 (+97%) | 1.89 (+101%) | 1.91 (+99%)
json_i8_as_i32 | 3.08 | 2.74 (+13%) | 2.66 (+16%) | 2.08 (+48%) | 2.10 (+47%) | 2.11 (+46%)
json_u8_as_u64 | 3.81 | 6.11 (-38%) | 3.97 (-4%) | 3.65 (+4%) | 3.62 (+5%) | 3.62 (+5%)
json_i8_as_i64 | 3.15 | 6.26 (-50%) | 4.31 (-27%) | 3.87 (-19%) | 3.86 (-19%) | 3.86 (-19%)
log_uniform_u32 | 8.83 | 2.49 (+255%) | 2.10 (+320%) | 1.66 (+431%) | 1.67 (+430%) | 1.67 (+428%)
log_uniform_u64 | 11.15 | 5.59 (+99%) | 3.55 (+214%) | 3.32 (+236%) | 3.28 (+240%) | 3.33 (+235%)
log_uniform_i32 | 7.83 | 2.50 (+213%) | 2.26 (+247%) | 1.82 (+331%) | 1.81 (+332%) | 1.81 (+333%)
log_uniform_i64 | 11.43 | 5.80 (+97%) | 3.84 (+198%) | 3.50 (+226%) | 3.50 (+227%) | 3.55 (+222%)
log_uniform_u128 | 77.85 | 10.76 (+623%) | 7.51 (+937%) | 6.84 (+1039%) | 6.73 (+1056%) | 6.73 (+1057%)
log_uniform_i128 | 53.63 | 10.20 (+426%) | 7.99 (+571%) | 8.23 (+551%) | 7.75 (+592%) | 7.75 (+592%)

</details>

I mentioned Champagne and Lemire and their AVX512 implementation above. I have not yet included AVX512 code for two reasons: 1. it is a wide field, there are so many variations of instructions available and different CPUs implement them with different characteristics, so it becomes hard to settle on one or two variants. Also, I mentioned above that CL use branches, and their data sets are not actually large enough to beat the branch predictor, so I would also have to demonstrate that I didn't tune my benchmarks just to beat them. All this seems to be too much work, especially given that distributions still don't push their defaults past v3 microarchitectures, and as I learned a few days ago, there really are [users that are bound to v1 for years to come](https://www.phoronix.com/news/CERN-Goes-Debian-Leaving-RHEL) and who are dear to my heart given that I did my PhD at CERN. This actually triggered the last commit which gained almost 25% on the SSE2 path.

One aside: I didn't use fmt's `format_int` because it has one defect that killed its usability: the output buffer has no padding. So It is not possible to copy the result out with a copy operation of predetermined length but instead a variable-length `memcpy` is needed, which introduces a second hard-to-predict branch on top of the loop over digits.

I should mention that the ARM {u,i}32 implementation is a bit over-optimized: M5 decodes 8 instructions at a time, so making the code aligned and < 32 instructions allows a benchmark to process it in 4 cycles. For this reason, and only this reason there is a variant of the zeros array that includes the leading '-' (that allowed shaving off one instruction at the expense of 16bytes more table data, and the re-arranging of the static_data: ARM can encode small offsets efficiently, so it is carefully layed out to make the relevant constants (mostly for the digit_count) available near the beginning of the struct with some dirty offset tricks. Given how short this function is, it may be a better choice to actually make it inlinable. This was also the main driver to place this code in a separate file, because you had also rearranged `static_data` to use more efficient addressing on ARM. One could put the things integers need inside a nested struct and then get ideal addressing for both floats and integers, but it is micro-optimization churn that obscures what is actually going on, and also a continuous source of merge conflicts, whereas in this way AI can handle the folding back of the code easily.

As for how to merge it, as I said, I moved it to a separate file to be able to easier review it, but I see not fundamental reason to not merge it back into the original file.

### DIFF
--- a/test/itoa-benchmark.cc
+++ b/test/itoa-benchmark.cc
@@ -1,0 +1,417 @@
+// Integer-formatting benchmarks for https://github.com/vitaut/zmij/,
+// comparing zmij against fmt (fmt::format_int, its fastest integer path).
+//
+// Three workloads:
+//  - json_u8/i8_*: serialize an array of byte-valued numbers ([0, 255] as
+//    u32/u64, [-128, 127] as i32/i64) as a JSON array "[ 123, 124, 0, -14 ]".
+//  - json_twitter_mk2_*: the same JSON-array serialization for a u64 stream
+//    whose digit-count sequence comes from a second-order Markov chain
+//    fitted on the integers of simdjson's twitter.json (see twitter_digits).
+//    The zmij32 variant dispatches to the u32 writer for values that fit,
+//    a user-side wrapper over the public API.
+//  - log_uniform_*: plain conversion of values log-uniformly distributed
+//    over the full range of the type (the same number of entries per digit
+//    count) for i32, i64 and i128.
+//
+// Copyright (c) 2025 - present, Victor Zverovich
+// Distributed under the MIT license (see LICENSE).
+
+#include <benchmark/benchmark.h>
+#include <stdint.h>  // uint64_t
+#include <string.h>  // memcpy
+
+#include <algorithm>  // std::shuffle
+#include <random>     // std::mt19937_64
+#include <vector>     // std::vector
+
+#include "fmt/format.h"
+#include "zmij.h"
+
+namespace {
+
+#if ZMIJ_USE_INT128
+using wide_uint = unsigned __int128;
+#else
+using wide_uint = uint64_t;
+#endif
+
+// Stream lengths. Long enough that the branch predictor cannot memorize
+// the per-iteration branch-outcome sequence across benchmark iterations
+// (on Zen 5 a replayed 64k-item stream is still ~96% memorized; miss
+// rates only saturate past ~128k items).
+constexpr size_t json_array_size = 1 << 20;
+constexpr size_t entries_per_digit_count = 16384;
+
+template <typename U> auto random_bits(std::mt19937_64& gen) -> U {
+  if constexpr (sizeof(U) > 8)
+    return (U(gen()) << 64) | gen();
+  else
+    return U(gen());
+}
+
+// Byte-valued numbers equidistributed over [lo, hi], stored as T.
+template <typename T> auto byte_range_data(int lo, int hi) -> std::vector<T> {
+  std::mt19937_64 gen(7);
+  std::uniform_int_distribution<int> dist(lo, hi);
+  std::vector<T> data(json_array_size);
+  for (auto& value : data) value = T(dist(gen));
+  return data;
+}
+
+// Log-uniform over the full range of T: the same number of entries for each
+// magnitude digit count, uniform within a digit count, random sign for
+// signed types.
+template <typename T> auto log_uniform_data() -> std::vector<T> {
+  constexpr bool is_signed = T(-1) < T(0);
+  const wide_uint max_magnitude =
+      is_signed ? (wide_uint(1) << (sizeof(T) * 8 - 1)) - 1
+      : sizeof(T) == sizeof(wide_uint)
+          ? ~wide_uint(0)
+          : (wide_uint(1) << (sizeof(T) * 8 % (sizeof(wide_uint) * 8))) - 1;
+  std::mt19937_64 gen(42);
+  std::vector<T> data;
+  wide_uint lo = 0, hi = 9;
+  for (;;) {
+    if (hi > max_magnitude) hi = max_magnitude;
+    wide_uint span = hi - lo + 1;  // 0 only for the full wide_uint range
+    for (size_t i = 0; i != entries_per_digit_count; ++i) {
+      wide_uint bits = random_bits<wide_uint>(gen);
+      T value = T(lo + (span != 0 ? bits % span : bits));
+      if (is_signed && (gen() & 1) != 0) value = T(0) - value;
+      data.push_back(value);
+    }
+    if (hi == max_magnitude) break;
+    lo = hi + 1;
+    hi = hi * 10 + 9;
+  }
+  std::shuffle(data.begin(), data.end(), gen);
+  return data;
+}
+
+// Digit-count sequence of the 2108 integers in simdjson's twitter.json, in
+// document order (status ids, user ids, counts, ...). Basis for value streams
+// whose digit-count sequence has twitter's structure: order 0 draws i.i.d.
+// from the marginal distribution, orders 1/2 from Markov chains fitted on the
+// circular sequence (circularity guarantees every reachable context was
+// observed). All three share the same marginal, so timing differences between
+// them isolate the effect of sequence predictability on the writers.
+const unsigned char twitter_digits[] = {
+    18,9,10,3,3,1,3,4,1,1,9,1,1,18,9,1,2,2,3,1,4,5,18,8,2,2,2,3,4,3,2,4,5,5,2,2,18,2,2,3,
+    3,3,3,3,3,3,3,2,1,8,1,2,18,2,2,3,3,3,3,3,3,3,3,18,18,18,9,9,1,2,4,3,2,3,5,5,1,1,9,1,
+    2,18,9,1,2,2,2,4,4,2,4,5,6,18,9,3,3,2,4,5,5,2,2,2,1,9,1,2,18,9,3,3,1,4,5,5,18,8,3,3,
+    2,4,5,5,4,4,1,2,18,2,2,3,3,4,3,3,3,3,3,4,1,2,2,8,1,2,18,2,2,3,3,4,3,3,3,3,3,18,18,10,
+    3,3,1,2,4,1,1,18,9,3,3,1,4,5,5,1,1,18,18,9,10,2,2,1,3,5,3,1,1,9,1,2,18,10,3,3,1,3,4,18,
+    10,10,3,3,1,4,4,1,1,10,1,2,1,1,10,1,2,10,2,2,18,10,1,1,1,1,4,1,1,18,10,3,4,1,1,3,18,10,3,3,
+    1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,3,1,3,5,18,18,
+    9,9,1,2,4,3,3,2,5,5,1,1,9,2,2,9,2,2,18,2,2,3,3,3,3,4,3,3,3,18,1,1,9,1,2,9,2,2,9,2,
+    2,18,2,3,3,3,3,3,4,3,3,3,18,18,10,3,3,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,4,4,2,4,
+    4,18,10,1,2,3,3,2,4,5,5,2,2,2,2,2,1,2,2,10,1,2,18,10,3,3,1,4,5,4,1,1,18,10,3,4,1,1,3,18,
+    10,3,3,1,1,3,2,1,2,1,10,1,2,18,9,4,4,2,4,5,6,18,10,1,2,3,3,4,4,2,3,5,5,1,1,3,3,1,1,3,
+    3,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,3,1,1,3,18,10,3,3,1,1,3,2,
+    1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,2,3,1,1,3,18,10,3,3,1,1,
+    3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,
+    1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,9,1,2,3,3,2,4,5,
+    6,18,9,1,2,2,2,6,2,4,2,5,5,3,3,3,1,9,1,2,18,10,2,3,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,
+    2,18,10,3,3,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,2,3,1,1,3,18,10,3,3,1,1,3,2,1,2,1,
+    10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,10,3,3,1,3,3,1,1,3,3,10,1,2,18,
+    10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,2,2,1,1,5,1,1,18,10,3,3,1,1,3,18,10,3,3,
+    1,1,3,2,1,2,1,10,1,2,18,10,3,3,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,3,1,1,3,18,10,
+    3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,2,2,1,2,3,
+    18,9,3,3,1,3,4,2,1,2,2,2,1,2,2,9,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,
+    10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,
+    2,18,10,3,3,1,3,4,1,1,18,10,1,2,1,2,1,1,6,1,1,2,3,2,2,18,3,3,3,3,3,3,3,3,3,3,18,10,2,3,
+    1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,9,3,3,1,4,5,5,1,1,18,8,1,2,2,3,3,3,2,5,5,6,18,
+    10,1,2,3,3,1,2,3,1,1,1,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,
+    1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,3,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,
+    3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,3,2,5,4,18,10,1,2,3,3,2,4,5,5,1,1,1,1,
+    10,1,2,18,10,3,3,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,
+    2,1,10,1,2,18,10,4,4,2,1,5,1,1,18,10,2,3,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,
+    1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,8,2,
+    2,3,3,1,4,5,6,18,18,9,8,4,3,3,5,5,5,1,1,2,2,2,3,9,2,2,1,1,2,2,3,3,8,1,2,9,2,2,18,10,
+    3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,3,1,1,4,5,1,1,3,3,18,18,9,9,1,2,3,3,1,
+    2,5,4,1,1,9,1,2,18,10,2,3,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,2,3,1,1,3,18,10,3,3,
+    1,1,3,2,1,2,1,10,1,2,18,10,3,3,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,2,2,1,1,5,1,1,
+    2,2,18,2,3,3,3,3,3,3,3,3,3,18,10,10,1,2,3,3,1,4,4,1,1,3,3,10,1,2,18,10,1,2,4,4,1,4,5,4,
+    1,1,18,10,1,1,1,1,4,1,1,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,
+    10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,
+    3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,3,1,3,4,1,1,9,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,
+    3,2,1,2,1,10,1,2,18,10,2,3,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,3,1,1,3,18,10,3,3,
+    1,1,3,2,1,2,1,10,1,2,18,10,2,3,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,
+    3,3,1,1,3,2,1,2,1,10,1,2,18,10,2,3,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,2,3,1,1,3,
+    18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,18,9,10,3,3,1,4,4,1,1,9,1,2,18,10,3,4,1,1,3,18,10,3,3,1,
+    1,3,2,1,2,1,10,1,2,18,18,9,10,2,2,1,5,5,5,1,1,9,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,
+    2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,
+    2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,3,1,1,3,18,10,3,3,1,
+    1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,
+    3,1,1,3,2,1,2,1,10,1,2,18,10,5,5,2,4,5,5,1,1,2,2,2,2,2,2,18,9,4,2,2,1,5,5,1,1,3,3,18,
+    10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,2,18,10,3,4,1,1,3,18,10,3,3,1,1,3,2,1,2,1,10,1,
+    2,18,18,10,10,2,2,1,3,3,1,1,10,1,2,18,10,1,2,3,3,1,1,5,5,1,1,3,3,18,9,3,3,1,4,5,18,10,4,4,
+    1,1,4,1,1,1,1,10,1,2,18,10,1,2,1,3,5,3,18,10,1,2,3,3,2,4,5,5,2,2,2,2,2,1,2,2,10,1,2,18,
+    9,3,2,1,2,5,18,8,4,3,2,3,5,5,1,1,2,2,18,2,3,3,3,3,3,3,3,3,3,1,1,2,3,8,1,2,18,3,3,3,
+    3,3,3,3,3,3,3,18,18,10,1,2,3,3,3,3,2,3,4,1,1,2,2,2,2,18,3,1
+};
+
+template <int order> auto twitter_markov_data() -> std::vector<uint64_t> {
+  constexpr int num_states = 21;  // digit counts 1..20, indexed directly
+  const int n = int(sizeof(twitter_digits));
+  std::vector<uint32_t> c0(num_states, 0);
+  std::vector<uint32_t> c1(num_states * num_states, 0);
+  std::vector<uint32_t> c2(num_states * num_states * num_states, 0);
+  for (int i = 0; i < n; ++i) {
+    int a = twitter_digits[i];
+    int b = twitter_digits[(i + 1) % n];
+    int c = twitter_digits[(i + 2) % n];
+    ++c0[a];
+    ++c1[a * num_states + b];
+    ++c2[(a * num_states + b) * num_states + c];
+  }
+  uint64_t pow10[20] = {1};
+  for (int i = 1; i < 20; ++i) pow10[i] = pow10[i - 1] * 10;
+
+  std::mt19937_64 gen(2108);
+  auto pick = [&gen](const uint32_t* row, uint32_t total) {
+    uint64_t r = gen() % total;
+    for (int s = 0;; ++s) {
+      if (r < row[s]) return s;
+      r -= row[s];
+    }
+  };
+  std::vector<uint64_t> data(json_array_size);
+  int a = twitter_digits[0], b = twitter_digits[1];
+  for (auto& value : data) {
+    int d = order == 0 ? pick(c0.data(), uint32_t(n))
+            : order == 1
+                ? pick(&c1[b * num_states], c0[b])
+                : pick(&c2[(a * num_states + b) * num_states],
+                       c1[a * num_states + b]);
+    a = b;
+    b = d;
+    uint64_t lo = d == 1 ? 0 : pow10[d - 1];
+    value = lo + gen() % (pow10[d] - lo);
+  }
+  return data;
+}
+
+struct zmij_write {
+  template <typename T> auto operator()(char* out, T value) const -> char* {
+    // n = 64 >= every integer buffer size, so this takes the direct path.
+    return zmij::write(out, 64, value);
+  }
+};
+
+// Like zmij_write, but routes 64-bit values that fit in 32 bits to the u32
+// kernel. The range check is a data-dependent but (on realistic streams)
+// highly predictable branch; values on the fast path skip the u64 kernel's
+// fixed 16-digit body.
+struct zmij_write32 {
+  auto operator()(char* out, uint64_t value) const -> char* {
+    if (value <= UINT32_MAX) return zmij::write(out, 64, uint32_t(value));
+    return zmij::write(out, 64, value);
+  }
+  auto operator()(char* out, int64_t value) const -> char* {
+    if (value == int64_t(int32_t(value)))
+      return zmij::write(out, 64, int32_t(value));
+    return zmij::write(out, 64, value);
+  }
+};
+
+// For 32/64-bit we call fmt::detail::format_decimal directly, the same
+// digit-writing primitive fmt::format_int wraps. It writes straight into the
+// output buffer with no intermediate copy, but requires the digit count up
+// front (count_digits), since it fills back-to-front within [out, out + n).
+// __int128 has no format_decimal, so 128-bit goes through fmt::format_to.
+struct fmt_write {
+  template <typename T> auto operator()(char* out, T value) const -> char* {
+    if constexpr (sizeof(T) > 8) {
+      return fmt::format_to(out, "{}", value);
+    } else if constexpr (T(-1) < T(0)) {
+      auto abs_value = static_cast<fmt::detail::uint32_or_64_or_128_t<T>>(value);
+      *out = '-';
+      if (value < 0) {
+        ++out;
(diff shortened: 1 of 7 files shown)


TobiSchluter:
I forgot to say two things:
1. this also adds tests, an exhaustive test of int32 as well as a test of all boundaries (powers of two and ten as well as their neighbors, numerical limits)
2. why is u128 not branch-free? Because the added work for the very long numbers takes similar time to a mispredicted branch, and because it is hard to imagine a scenario where the user's data would touch the whole range of possible u128. Of course, if their numbers keep straddling the 2^64 boundary in unpredictable ways, this will pessimize the performance, but that seems like a low risk especially given that it is still much faster than fmt's code in that range.

zmij-int.cc
```diff
@@ -0,0 +1,1589 @@
...
+  // also biasing by '0', so the low byte of each lane is directly the ASCII code
+  // -- no cvttps and no separate '0' add. The add runs parallel with the shift.
+  const __m128 bias = _mm_load_ps(d.peel4_bias);
+  __m128 xf = _mm_cvtepi32_ps(_mm_set1_epi32(int(n)));
+  __m128 qf = _mm_round_ps(_mm_mul_ps(xf, recip),
+                           _MM_FROUND_TO_ZERO | _MM_FROUND_NO_EXC);
+  __m128 shifted = _mm_castsi128_ps(_mm_slli_si128(_mm_castps_si128(qf), 4));
+  // digit_k + '0' + 2^23 = (qf_k + bias) - 10 * qf_{k-1}
+  __m128i dig =
+      _mm_castps_si128(_mm_fnmadd_ps(ten, shifted, _mm_add_ps(qf, bias)));
```
vitaut, line 743: This need to check whether FMA is enabled.

zmij-int.cc
```diff
@@ -0,0 +1,1589 @@
...
+    copy_bytes(buf + 24, &low_hi, 8);
+    copy_bytes(buf + 32, &low_lo, 8);
+    if (!big) {  // 19-32 digits: top (<=16) + low 16
+      uint64_t top = uint64_t(lo.quot);
+      uint64_t top_hi = to_bcd8(uint32_t(top / 100'000'000ull)).bcd + zeros;
+      uint64_t top_lo = to_bcd8(uint32_t(top % 100'000'000ull)).bcd + zeros;
+      copy_bytes(buf + 8, &top_hi, 8);
+      copy_bytes(buf + 16, &top_lo, 8);
+      uint64_t len = 16 + count_digits(top, *d);
+      copy_bytes(out, buf + 40 - len, 32);
```
vitaut, line 1455: This may copy uninitialized data from stack into the output buffer. Not sure if it matters but maybe it's worth zero initializing buf if it doesn't affect perf?

test/itoa-benchmark.cc
```diff
@@ -0,0 +1,417 @@
...
+  std::uniform_int_distribution<int> dist(lo, hi);
+  std::vector<T> data(json_array_size);
+  for (auto& value : data) value = T(dist(gen));
+  return data;
+}
+
+// Log-uniform over the full range of T: the same number of entries for each
+// magnitude digit count, uniform within a digit count, random sign for
+// signed types.
+template <typename T> auto log_uniform_data() -> std::vector<T> {
```
vitaut, line 64: log_uniform_u128 is not log-uniform over the 39 decimal digit counts because hi = hi * 10 + 9 overflows before it can be clamped to UINT128_MAX.

test/zmij-itoa-test.cc
```diff
@@ -0,0 +1,139 @@
...
+// Maximum value of T as an unsigned magnitude.
+template <typename T> constexpr auto max_magnitude() -> wide_uint {
+  return is_signed_int<T>()
+             ? (wide_uint(1) << (sizeof(T) * 8 - 1)) - 1
+             : (sizeof(T) == sizeof(wide_uint)
+                    ? ~wide_uint(0)
+                    : (wide_uint(1) << (sizeof(T) * 8 % (sizeof(wide_uint) * 8))) - 1);
+}
+
+template <typename T> constexpr auto buffer_size() -> size_t {
```
vitaut, line 43: This gives one extra char for unsigned.

zmij-int.cc
```diff
@@ -0,0 +1,1589 @@
...
+#include <stdint.h>  // uint64_t
+#include <string.h>  // memcpy
+
+#include <type_traits>  // std::make_unsigned, std::conditional_t
+
+// The configuration macros below mirror zmij.cc so that both files compile
+// standalone with the same settings; identical redefinitions keep unity
+// builds that include both files valid.
+
+#ifndef ZMIJ_USE_SIMD
```
vitaut, line 24: I think we should put common configuration macros in `zmij.h` to avoid duplication.

vitaut (changes requested):
Impressive work!

Overall looks good but please address inline comments. Also I would drop the exhaustive test. It is useful for float because of hard-to-test boundary conditions but I don't think it's particularly useful for integers.

zmij-int.cc
TobiSchluter: This block is conditional  on `ZMIJ_USE_AVX2_U64_FP` which in turn checks for  `ZMIJ_USE_AVX2 && (defined(__znver5__) || defined(__tune_znver5__))`.

Other than that, I just learned from StackOverflow that there is a CPU out in the wild that has AVX2 but not FMA, which seems like a mistake. So, I don't think a local check is mandated, but there might be a CPU with FMA and no AVX2 (and lots of hardware multipliers) that could benefit from this?

zmij-int.cc
TobiSchluter: I intentionally omitted the initialization and added the comment on line 1441ff after the initialization did indeed show up in benchamrks.  I went through a few iterations of what goes where in the buffer and when but it's a large search-space, and so there may be a faster variation with initialization.

But then OTOH: the zero initialization is much cheaper with SIMD instructions. Hence the compiler will use them when available no matter what our preprocessor variables say. All of this is a bit academic for this fallback for ... I don't know ... PowerPC? RISC-V? Ancient Android phones? What is the actual target group?

zmij-int.cc
TobiSchluter: Oh, I think I get it now: if the user selects an `-march` that enables AVX2 but not FMA they might see an error if they also enable `ZMIJ_USE_AVX2_U64_FP`. Gotta love the randomness of the x86 instruction set growth! Ok, I will make sure that the checks are tight.

More as a note to self, this also applies to `itoa_top8` below (which is I think the most creative thing I did in all of this, and which I failed to highlight in the PR).

MasterDuke17:
Would you be able to also make a C version of this? MoarVM uses the C floating point code, and I would love to use Żmij for integers as well.

TobiSchluter:
Since my Zen5 broke, I moved to an i9 (Raptor Lake). Great to physically experience a modern Intel CPU that doesn't have AVX512. Performance is distributed a bit differently -- sign handling seems to cost more. I added the benchmark table to the first post.

One curious observation that I made because I inadvertently ran the floating-point benchmarks as well is that for `ftoa` on the i9 the non-SIMD variant is actually fastest.

zmij-int.cc
vitaut: > if the user selects an -march that enables AVX2 but not FMA they might see an error 

Yes, that was my concern.

TobiSchluter: I tried to address this concern with some infrastructure in #160

test/itoa-benchmark.cc
TobiSchluter: Indeed. This led to 39 digits being overpopulated in the u128 benchmarks. A before and after with the fixed 00d791a41673cb7b98b1c99b88ddd6f67336ad0c is here, i128 is unchanged up to measurement noise:

## Intel i7-1165G7 (Tiger Lake), gcc 16

| benchmark | fmt | scalar | v1 | v2 | v3 | native |
|---|---|---|---|---|---|---|
| log_uniform_u128 (before) | 215.12 | 22.63 (+850%) | 18.76 (+1047%) | 16.77 (+1183%) | 19.34 (+1012%) | 16.42 (+1210%) |
| log_uniform_u128 (after) | 135.13 | 19.55 (+591%) | 16.32 (+728%) | 16.37 (+725%) | 15.40 (+778%) | 15.47 (+773%) |
| log_uniform_i128 (before) | 134.14 | 21.80 (+515%) | 18.77 (+615%) | 18.07 (+643%) | 18.02 (+644%) | 17.61 (+662%) |
| log_uniform_i128 (after) | 147.42 | 21.70 (+579%) | 18.56 (+694%) | 19.91 (+640%) | 19.11 (+671%) | 25.08 (+488%) |

## Intel i7-1165G7 (Tiger Lake), clang 21

| benchmark | fmt | scalar | v1 | v2 | v3 | native |
|---|---|---|---|---|---|---|
| log_uniform_u128 (before) | 154.24 | 28.59 (+439%) | 16.27 (+848%) | 16.35 (+844%) | 15.13 (+920%) | 15.51 (+894%) |
| log_uniform_u128 (after) | 112.93 | 20.89 (+441%) | 15.20 (+643%) | 15.64 (+622%) | 14.44 (+682%) | 18.47 (+511%) |
| log_uniform_i128 (before) | 108.61 | 21.65 (+402%) | 18.26 (+495%) | 20.03 (+442%) | 15.82 (+586%) | 15.85 (+585%) |
| log_uniform_i128 (after) | 106.01 | 21.16 (+401%) | 18.54 (+472%) | 18.73 (+466%) | 16.49 (+543%) | 16.30 (+550%) |

test/zmij-itoa-test.cc
TobiSchluter: Thanks for catching that. Fixed in 37a607482bff5cc82e2ced82614675c431cce7e3

PR #160, opened 2026-09-17
Add more expressive ZMIJ_AMD64_LEVEL variable.

TobiSchluter:
Allows to encode more than just SSE2, SSE4.1. Mainly for the benefit of itoa which brings AVX2 variants.

This is mainly driven by @vitaut's question here https://github.com/vitaut/zmij/pull/156#discussion_r4027660857 If the user can input random combinations of x64 features it becomes unwieldy to check every site against all possibilities.

I kept the SSE and SSE4_1 macros as backwards-compatible user-interface. They are also still used in the tests.

Built and tested across the levels, also inspected that the various combinations come out correctly.

ps I was missing a few checks in the first commit. I think it's not worth identifying the exact combinations of flags that will work with the code and only require those features, instead the code now hews closely to the microarchitecture level definitions.


vitaut (changes requested):
I think feature-specific macros such as `ZMIJ_USE_SSE`, `ZMIJ_USE_SSE4_1` and `ZMIJ_USE_AVX2` are more intuitive than an ordinal `ZMIJ_AMD64_LEVEL`.

The level also bundles together capabilities that are not necessarily required by the same code. For example, code using AVX2 shouldn't implicitly require FMA, BMI2, MOVBE, etc. I'd prefer extending the existing `ZMIJ_USE_*` scheme and checking additional features such as FMA separately where they are actually needed.

TobiSchluter:
I thought about this a little, so please excuse the late answer. What I was trying to do was to formalize and extend what we are doing already: the SSE 4.1 flag enables v2-level instructions -- we don't add the shuffles from SSSE3 as a SSSE3 variety and then enable another set of shuffle constants once we add SSE 4.1's mullo32.

I'm also seeing the millions of permutations this can create, especially given that the grouping in features is fairly random, not really reflective of real CPUs (AVX2, say, always comes with FMA3 (with the exception of some obscure VIA CPUs)) and not monotonous (this mostly affects v4 and beyond, but then there's e.g. IFMA backported to VEX encoding, i.e. the AVX2 world). At the same time, compilers on most (all?) distributions default to one of v1, v2, v3, and this is what the code is going to be built against.

Anyway, I'll go ahead with the pure feature-based enabling/disabling, and see if it turns into something beautiful.

That said, I have variants of jeaiii that are faster on Intel for `ftoa` and `itoa32` than the SIMD code, so at least there the number of variants might shrink. The `ftoa` may be obsolete or incompatible with your latest changes, though. The `ftoa` version is here https://github.com/TobiSchluter/zmij/commit/a4b448482b92247a7a6f34dc9e34336af9a0d99c

